PR #7, opened 2019-06-30
Check for config daemon locally first

Mmduh-483:


Mmduh-483:
@pliurh after this patch merged, please update the image because of `bindata` dir is already inside the image.

PR #8, opened 2019-07-02
Use openshift origin image

pliurh:


deploy/operator.yaml
```diff
@@ -22,7 +22,7 @@ spec:
       containers:
         - name: sriov-network-operator
           # Replace this with the built image name
-          image: quay.io/pliurh/sriov-network-operator
+          image: quay.io/openshift/origin-sriov-network-operator
```
zshi-redhat: do we need to add version as image suffix? for example `quay.io/openshift/origin-sriov-network-operator:4.2`

pliurh: By default, it will pull the `latest` image, which I think it is fine. Since this file is mainly for developers and upstream. For downstream, the operator is deployed though CVS files.

zshi-redhat: yes, it should have no real effect on the image pulling, perhaps just for documentation: https://github.com/openshift/cluster-network-operator/commit/58cea9936c9db814eb658bf444afb8eb66dda675

/lgtm

PR #16, opened 2019-07-10
sriovnetwork crd: add vf properities

zshi-redhat:
- spoof check
- trusted vf


pkg/apis/sriovnetwork/v1/sriovnetwork_types.go
```diff
@@ -16,6 +16,8 @@ type SriovNetworkSpec struct {
 	// +kubebuilder:validation:Minimum=0
 	// +kubebuilder:validation:Maximum=4096
 	Vlan             int    `json:"vlan,omitempty"`
+	Spoof            string `json:"spoof,omitempty"`
```
moshe010: why not a bool?

pkg/apis/sriovnetwork/v1/sriovnetwork_types.go
```diff
@@ -16,6 +16,8 @@ type SriovNetworkSpec struct {
...
+	Trust            string `json:"trust,omitempty"`
 }
```
moshe010: maybe we should add state { auto | enable | disable} ?
- auto: a reflection of the PF link state (default)
- enable: lets the VF to communicate with other VFs on this host even if the PF link state is down
- disable: causes the HW to drop any packets sent by the VF.

pkg/apis/sriovnetwork/v1/sriovnetwork_types.go
pliurh: @zshi-redhat to add any new field in CRD, you have to run `make gencode` to update the CRD template and deep-copy. You may also want to check https://github.com/operator-framework/operator-sdk/blob/master/doc/user-guide.md#add-a-new-custom-resource-definition. 

And there is issue of operator-sdk. So after running `make gencode`. Please make sure only the relevant files are updated, and the API group in CRD manifest is unchanged.

pkg/apis/sriovnetwork/v1/sriovnetwork_types.go
zshi-redhat: This is a good one. I'm going to add parameters for what will be supported by SRIOV-CNI in this PR: https://github.com/intel/sriov-cni/pull/92

pkg/apis/sriovnetwork/v1/sriovnetwork_types.go
zshi-redhat: Updated to use `bool` type

zshi-redhat: Done

pliurh:
Please also update the CRD in mainifests directory and the example in clusterversionservice file.

pkg/apis/sriovnetwork/v1/sriovnetwork_types.go
pliurh: I think it may be better to use value instead of pointer here. So that develop can use this struct without considering deepcopy problem.

zshi-redhat: This was originally a value, I changed it due to that I'd like to differentiate between unset and zero values of boolean type, only insert the config in CNI once it's not nil.

PR #20, opened 2019-07-12
Load RDMA kernel modules for RHEL node

pliurh:


pliurh:
@Mmduh-483 @moshe010 Please take a look at this patch. It will install and active rdma service on all non-RHCOS worker nodes. I am not sure if there is any side effect when a node is not running rdma application but with rdma kmods loaded. Please let me know your thought on this. And if you don't mind, could you test this patch in your environment?

bindata/scripts/enable-rdma.sh
```diff
@@ -0,0 +1,55 @@
...
+else
+  if [ -f /host/usr/libexec/rdma-init-kernel ]; then
+    echo "rdma.service installed"
+    if kmod_isloaded; then
+      exit
+    else
+      trigger_udev_event
+    fi
+  else
+    yum install -y rdma-core
```
zshi-redhat: Q: this is executed in sriov config daemon, will the pkg be installed on the host? do we need to `chroot` ?

bindata/scripts/enable-rdma.sh
```diff
@@ -0,0 +1,55 @@
...
+    true
+  else
+    echo "RDMA kernel modules not loaded"
+    false
+  fi
+}
+
+function trigger_udev_event {
+  echo "Trigger udev event"
+  chroot /host/ modprobe -r mlx4_en && chroot /host/ modprobe mlx4_en
```
zshi-redhat: This is dangerous, if the default interface on the host is using mlx4_en or mlx5_core driver, then the worker node will lose connection with openshift masters.

bindata/scripts/enable-rdma.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
+
+REDHAT_RELEASE_FILE="/etc/redhat-release"
+
+function kmod_isloaded {
+  if grep --quiet '\(^ib\|^rdma\)' <(chroot /host/ lsmod); then
```
zshi-redhat: nit: can we list all the module names instead of using regex?

bindata/scripts/enable-rdma.sh
```diff
@@ -0,0 +1,55 @@
...
+  echo "Trigger udev event"
+  chroot /host/ modprobe -r mlx4_en && chroot /host/ modprobe mlx4_en
+  chroot /host/ modprobe -r mlx5_core && chroot /host/ modprobe mlx5_core
+}
+
+if [ ! -f "$REDHAT_RELEASE_FILE" ]; then
+  exit 1
+fi
+
+if ! grep --quiet 'mlx4_en' <(chroot /host/ lsmod) && ! grep --quiet 'mlx5_core' <(chroot /host/ lsmod); then
```
zshi-redhat: Q: Are mlx drivers loaded by default? is there any extra config required?

bindata/scripts/enable-rdma.sh
pliurh: Fixed

bindata/scripts/enable-rdma.sh
pliurh: hmm, that is a problem. However, since connect will go back immediately and this activity will only be triggered one time, I think that is acceptable.

bindata/scripts/enable-rdma.sh
pliurh: I think those drivers will be loaded automatically if there is any such device detected. @Mmduh-483 @moshe010  Please correct me if I am wrong.

bindata/scripts/enable-rdma.sh
pliurh: The loaded module names could be different with different hardware. This line only tends to check if any of the rdma kernel modules has been loaded, if any of the kmod is loaded, that means the rdma service has been activated.

pkg/daemon/daemon.go
```diff
@@ -76,6 +80,8 @@ func (dn *Daemon) Run(stopCh <-chan struct{}, exitCh <-chan error) error {
 	glog.V(0).Info("Run(): start daemon")
 	// Only watch own SriovNetworkNodeState CR
 
+	tryEnableRdma()
```
Mmduh-483: There is no check if it failed or passed?

pkg/daemon/daemon.go
```diff
@@ -51,6 +54,7 @@ type Daemon struct {
 	mu *sync.Mutex
 }
 
+const scriptsPath = "/bindata/scripts/enable-rdma.sh"
```
Mmduh-483: can you change the cont name to something like `EnableMlnxRdmaScript`

moshe010: we don't need to change it, because RDMA is not just for mellanox. It can work with other vandors

bindata/scripts/enable-rdma.sh
moshe010: @pliurh, yes you are right. rdma-core provide the following udev rules https://github.com/linux-rdma/rdma-core/blob/master/kernel-boot/rdma-hw-modules.rules which should load all the drivers for the detected device

pkg/daemon/daemon.go
pliurh: If it failed, it just means this host is RDMA incapable, this daemon shall keep running. But maybe we shall add a field in status of SriovNetworkNodeState to indicate whether host is RDMA capable or not.

bindata/scripts/enable-rdma.sh
zshi-redhat: The link state can go back soon, not sure if all other network configuration on that interface can be properly restored.
if yes, we can fix it later.

PR #21, opened 2019-07-12
Do not hardcode GOOS/GOARCH

yselkowitz:
This is necessary for building on platforms other than amd64.


pliurh:
@yselkowitz Thanks for your patch. I have no problem to building code for other platform. But one thing I want to let you know is that. As you may know, there are 2 components in sriov-network-operator, the controller and the daemon. For controller, it has zero platform dependency. However for the daemon part, since it configs system/hardware related pieces, it is designed to run on linux amd64 only at least for now.

PR #23, opened 2019-07-14
Skip generic plugin if reboot needed

Mmduh-483:


pliurh:
@Mmduh-483 Have you tested loading vfio driver for Mellanox NICs with this patch?

pliurh:
@Mmduh-483  I think this patch will cause problem when generic plugin want to trigger reboot. I think we shall make this logic for Mellanox  nic only, like, if mlx plugin need a reboot then not executing Apply() of generic plugin.

PR #24, opened 2019-07-15
Add syncStatus and lastSyncError to SrivoNetworkNodeState CR

pliurh:


manifests/4.2/sriov-network-operator-sriovnetwork.crd.yaml
```diff
@@ -44,6 +40,8 @@ spec:
               type: boolean
             trust:
               type: boolean
+              minimum: 0
```
zshi-redhat: why adding min/max for boolean type value?

manifests/4.2/sriov-network-operator-sriovnetworknodestate.crd.yaml
```diff
@@ -29,6 +29,8 @@ spec:
           type: object
         spec:
           properties:
+            dpConfigVersion:
```
zshi-redhat: I didn't see this is used anywhere `dpConfigVersion`, what should be the value of it?

pliurh: It is not related to this patch, but a update to make sure the crds in `manifests` directory is the same as the ones in `deploy`

manifests/4.2/sriov-network-operator-sriovnetwork.crd.yaml
pliurh: It is a error caused by rebase, I'll fix it

PR #25, opened 2019-07-16
Add cluster roles required by the operator

booxter:
It needs the new roles for webhook deployment and other things.

PR #26, opened 2019-07-16
Fix RBAC for deploying cluster resource

pliurh:


pliurh:
fix https://github.com/openshift/sriov-network-operator/pull/25#issuecomment-511647120

PR #33, opened 2019-07-25
Fix e2e tests

pliurh:


hack/env.sh
```diff
@@ -1,8 +1,9 @@
...
-SRIOV_NETWORK_CONFIG_DAEMON_IMAGE=quay.io/openshift/origin-sriov-network-config-daemon
-RESOURCE_PREFIX=openshift.com
+export SRIOV_CNI_IMAGE=quay.io/openshift/origin-sriov-cni:4.2.0
+export SRIOV_DEVICE_PLUGIN_IMAGE=quay.io/openshift/origin-sriov-network-device-plugin:4.2.0
+export NETWORK_RESOURCES_INJECTOR_IMAGE=quay.io/openshift/origin-sriov-dp-admission-controller:4.2.0
+export RELEASE_VERSION=0.0.1-snapshot
+export NAMESPACE=sriov-network-operator
+export OPERATOR_NAME=sriov-network-operator
+export DAEMON_IMAGE_DIGEST=$(skopeo inspect docker://quay.io/openshift/origin-sriov-network-config-daemon | jq --raw-output '.Digest')
+export SRIOV_NETWORK_CONFIG_DAEMON_IMAGE=${SRIOV_NETWORK_CONFIG_DAEMON_IMAGE:-quay.io/openshift/origin-sriov-network-config-daemon@${DAEMON_IMAGE_DIGEST}}
```
zshi-redhat: what's the difference between using `DIGEST` and `4.2.0` ?

pkg/controller/sriovnetworknodepolicy/sriovnetworknodepolicy_controller.go
```diff
@@ -149,9 +145,15 @@ func (r *ReconcileSriovNetworkNodePolicy) Reconcile(request reconcile.Request) (
 	if err = r.syncConfigDaemonSet(defaultPolicy); err != nil {
 		return reconcile.Result{}, err
 	}
-	// Render and sync Daemon objects
-	if err = r.syncPluginDaemonObjs(defaultPolicy, policyList); err != nil {
-		return reconcile.Result{}, err
+	if len(policyList.Items) > 1 {
```
zshi-redhat: Is it an option to edit existing default policy CR and trigger provisioning of SRIOV vf?

test/e2e/e2e_test.go
```diff
@@ -97,7 +97,7 @@ func testWithOneSriovNetworkNodePolicyCR(t *testing.T, ctx *framework.TestCtx) {
 		Spec: sriovnetworkv1.SriovNetworkNodePolicySpec{
 			ResourceName: "resource-1",
 			NodeSelector: map[string]string{
-				"feature.node.kubernetes.io/sriov-capable": "true",
+				"feature.node.kubernetes.io/network-sriov.capable": "true",
```
zshi-redhat: I see we use `sriov-capable` in `Readme` and `manifest/4.2`, is there particular reason to change the label name?
shall we change them all ?

hack/env.sh
pliurh: The imagePullPolicy is ifNotPresent, so the latest image with 4.2.0 label can not be pulled when running test. Using digest to make sure the test will alway run against the latest origin image.

pkg/controller/sriovnetworknodepolicy/sriovnetworknodepolicy_controller.go
pliurh: No, default policy shall contain no configuration at all. Adding this condition to make sure the configure map is only created when there is at least one policy defined by user.

hack/env.sh
zshi-redhat: should the change be applied to all images including sriov-cni, devcie plugin and admission controller?

pkg/controller/sriovnetworknodepolicy/sriovnetworknodepolicy_controller.go
zshi-redhat: Do we need to include cni-daemonset and webhook in the same condition?

test/e2e/e2e_test.go
pliurh: This label name is from NFD https://github.com/kubernetes-sigs/node-feature-discovery. NFD changed the label name, so we changed accordingly.

Regarding to README file, I will make some change in anther PR.

hack/env.sh
pliurh: Good point. I'll add it for other images

pkg/controller/sriovnetworknodepolicy/sriovnetworknodepolicy_controller.go
pliurh: cni-daemonset has already been included. For webhook, I think we want it to be deployed regardless. So it doesn't need to care if there is any user defined policy or not.

hack/env.sh
```diff
@@ -1,9 +1,12 @@
-export SRIOV_CNI_IMAGE=quay.io/openshift/origin-sriov-cni:4.2.0
-export SRIOV_DEVICE_PLUGIN_IMAGE=quay.io/openshift/origin-sriov-network-device-plugin:4.2.0
-export NETWORK_RESOURCES_INJECTOR_IMAGE=quay.io/openshift/origin-sriov-dp-admission-controller:4.2.0
+CNI_IMAGE_DIGEST=$(skopeo inspect docker://quay.io/openshift/origin-sriov-cni | jq --raw-output '.Digest')
```
zshi-redhat: nit: This requires `skopeo` and `jq` be installed first, may be worth documenting.

PR #42, opened 2019-07-29
Fix the nic selection logic for updating SriovNetworkNodeState

pliurh:
When numVfs is not set, the device device ID of PF shall be used; otherwise, ID of VF shall be used.

### DIFF
--- a/pkg/apis/sriovnetwork/v1/helper.go
+++ b/pkg/apis/sriovnetwork/v1/helper.go
@@ -60,8 +60,8 @@ func UniqueAppend(inSlice []string, strings ...string) []string {
 
 // Apply policy to SriovNetworkNodeState CR
 func (p *SriovNetworkNodePolicy) Apply(state *SriovNetworkNodeState) {
-	s := p.Spec.NicSelector
-	if s.Vendor == "" && s.DeviceID == "" && len(s.RootDevices) == 0 && len(s.PfNames) == 0 {
+	s := p.Spec
+	if s.NicSelector.Vendor == "" && s.NicSelector.DeviceID == "" && len(s.NicSelector.RootDevices) == 0 && len(s.NicSelector.PfNames) == 0 {
 		// Empty NicSelector match none
 		return
 	}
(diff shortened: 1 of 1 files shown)


pkg/apis/sriovnetwork/v1/helper.go
```diff
@@ -97,8 +84,10 @@ func (s *SriovNetworkNicSelector) Selected(iface *InterfaceExt) bool {
 	if s.Vendor != "" && s.Vendor != iface.Vendor {
 		return false
 	}
-	if s.DeviceID != "" && s.DeviceID != iface.DeviceID {
-		return false
+	if s.DeviceID != "" {
+		if ((iface.NumVfs == 0 && s.DeviceID != iface.DeviceID) || (iface.NumVfs > 0 && s.DeviceID != SriovPfVfMap[iface.DeviceID])) {
```
zshi-redhat: Will this prevent user from resetting vf to 0 for a PF device?

pliurh: No, it will not. The function just help to make sure that only the NodeState CR which contains matched NICs will be configured.

zshi-redhat: when :
`s.DeviceID` is a VF id,
`iface.DeviceID` is a PF id,
`iface.NumVfs` is zero, (not configured with any VF yet)
Then:
Will the `iface` (PF) be ignored in merged config spec?

pliurh: Correct.

zshi-redhat: In such case, does it mean PF will not be configured even numvfs is specified in node policy because PF is not listed in config spec?

pliurh: The config in spec of NodeState is the output of policies. So if the numVfs is specified, it will compare the device ID with VF's, if match, then generate the config for the PF. Otherwise, it will compare the device ID with PF's.

pliurh: Yes, it's a problem. I've updated the code. Please take a look.

pkg/apis/sriovnetwork/v1/helper.go
```diff
@@ -80,19 +80,19 @@ func (p *SriovNetworkNodePolicy) Apply(state *SriovNetworkNodeState) {
...
-func (s *SriovNetworkNicSelector) Selected(iface *InterfaceExt) bool {
-	if s.Vendor != "" && s.Vendor != iface.Vendor {
+func (s *SriovNetworkNodePolicySpec) Selected(iface *InterfaceExt) bool {
+	if s.NicSelector.Vendor != "" && s.NicSelector.Vendor != iface.Vendor {
 		return false
 	}
-	if s.DeviceID != "" {
-		if ((iface.NumVfs == 0 && s.DeviceID != iface.DeviceID) || (iface.NumVfs > 0 && s.DeviceID != SriovPfVfMap[iface.DeviceID])) {
+	if s.NicSelector.DeviceID != "" {
+		if ((s.NumVfs == 0 && s.NicSelector.DeviceID != iface.DeviceID) || (s.NumVfs > 0 && s.NicSelector.DeviceID != SriovPfVfMap[iface.DeviceID])) {
```
zshi-redhat: If `s.NicSelector.DeviceID` is a VF ID and `iface.DeviceID` is equal to this VF ID, it seems that `iface` will be listed in the config spec, but it's not configurable because it's a VF device.

pliurh: The for loop in Apply() only walk through all the PFs. And in spec of NodeState, the desired configuration is indexed by PF only.

zshi-redhat: ok, thanks for the explanation, I think it should be good to go now.

PR #46, opened 2019-07-31
Use rpm-ostree to set kargs on RHCOS

pliurh:


zshi-redhat:
Shall SR-IOV Operator configure kargs via MCO instead of doing it directly?

pliurh:
There are some cons of using MCO approach:
1. This feature in MCO only works on RHCOS node. 
2. MCO can only apply same configuration to all the worker nodes indiscriminately. 
3. MCO will trigger node reboot after setting kargs, which may cause interference with our configure daemon.

zshi-redhat:
> There are some cons of using MCO approach:
> 
> 1. This feature in MCO only works on RHCOS node.
> 2. MCO can only apply same configuration to all the worker nodes indiscriminately.
The worker node can be selected by applying a label in `MachineConfig`, not necessary apply to all the worker nodes.

> 3. MCO will trigger node reboot after setting kargs, which may cause interference with our configure daemon.
I thought sriov operator also triggers reboot when kargs is configured, how that differs with MCO reboot.

pliurh:
> > There are some cons of using MCO approach:
> > 
> > 1. This feature in MCO only works on RHCOS node.
> > 2. MCO can only apply same configuration to all the worker nodes indiscriminately.
> 
> The worker node can be selected by applying a label in `MachineConfig`, not necessary apply to all the worker nodes.
That's true. However arbitrary label is not allowed. Basically, you can either use `machineconfiguration.openshift.io/role: worker` or `machineconfiguration.openshift.io/role: master`, which means it doesn't distinguish worker node.
> 
> > 1. MCO will trigger node reboot after setting kargs, which may cause interference with our configure daemon.
> 
> I thought sriov operator also triggers reboot when kargs is configured, how that differs with MCO reboot.
In config daemon, it will first send the desired state to each plugin, so all plugins can do some necessary job before reboot. If using MCO, updating kargs will automatically trigger reboot immediately, other plugin may not got the change to config the device before reboot.

zshi-redhat:
> > > There are some cons of using MCO approach:
> > > 
> > > 1. This feature in MCO only works on RHCOS node.
> > > 2. MCO can only apply same configuration to all the worker nodes indiscriminately.
> > 
> > 
> > The worker node can be selected by applying a label in `MachineConfig`, not necessary apply to all the worker nodes.
> 
> That's true. However arbitrary label is not allowed. Basically, you can either use `machineconfiguration.openshift.io/role: worker` or `machineconfiguration.openshift.io/role: master`, which means it doesn't distinguish worker nodes.

Got it, I didn't realized this was the case. Thanks for explanation.

> 
> > > 1. MCO will trigger node reboot after setting kargs, which may cause interference with our configure daemon.
> > 
> > 
> > I thought sriov operator also triggers reboot when kargs is configured, how that differs with MCO reboot.
> 
> In config daemon, it will first send the desired state to each plugin, so all plugins can do some necessary job before reboot. If using MCO, updating kargs will automatically trigger reboot immediately, other plugin may not got the change to config the device before reboot.

cgwalters:
Some sort of CC on pull requests like this to the machine-config-operator team would be a good idea.

pliurh:
@cgwalters Will do.

BTW, may I ask if there is any plan in MCD to support to config worker node in multiple groups, in other word, allow work nodes be configured differently?

PR #50, opened 2019-08-02
Remove existing net-att-def and create a new one when networkNamepace…

pliurh:
… in SriovNetwork CR changed

Fix https://bugzilla.redhat.com/show_bug.cgi?id=1735710

### DIFF
--- a/pkg/controller/sriovnetwork/sriovnetwork_controller.go
+++ b/pkg/controller/sriovnetwork/sriovnetwork_controller.go
@@ -93,6 +93,7 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 }
 
 var _ reconcile.Reconciler = &ReconcileSriovNetwork{}
+var lastNetworkNamespace = ""
 
 // ReconcileSriovNetwork reconciles a SriovNetwork object
 type ReconcileSriovNetwork struct {
@@ -115,6 +116,7 @@ func (r *ReconcileSriovNetwork) Reconcile(request reconcile.Request) (reconcile.
 
 	// Fetch the SriovNetwork instance
 	instance := &sriovnetworkv1.SriovNetwork{}
+	var namespaces = ""
 	err := r.client.Get(context.TODO(), request.NamespacedName, instance)
 	if err != nil {
 		if errors.IsNotFound(err) {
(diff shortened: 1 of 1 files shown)


pkg/controller/sriovnetwork/sriovnetwork_controller.go
```diff
@@ -141,33 +143,66 @@ func (r *ReconcileSriovNetwork) Reconcile(request reconcile.Request) (reconcile.
...
+				if err != nil {
+					reqLogger.Error(err, "Couldn't create NetworkAttachmentDefinition", "Namespace", netAttDef.Namespace, "Name", netAttDef.Name)
+					return reconcile.Result{}, err
+				}
+				lastNetworkNamespace = netAttDef.GetNamespace()
+			}
+			// NetworkAttachmentDefinition created successfully - don't requeue
+			return reconcile.Result{}, err
+		}
+		if lastNetworkNamespace != netAttDef.GetNamespace() {
```
zshi-redhat: Will this condition ever be `true`? It's in the loop of `lastNetworkNamespace == netAttDef.GetNamespace()`;  if netAttDef is created in the loop,  `lastNetworkNamespace` will be assigned with `netAttDef.GetNamespace()`.

pkg/controller/sriovnetwork/sriovnetwork_controller.go
```diff
@@ -93,6 +93,7 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 }
 
 var _ reconcile.Reconciler = &ReconcileSriovNetwork{}
+var lastNetworkNamespace = ""
```
zshi-redhat: Will `lastNetworkNamespace` var be overwritten if there are multiple SriovNetwork?

pkg/controller/sriovnetwork/sriovnetwork_controller.go
pliurh: Fixed.

pkg/controller/sriovnetwork/sriovnetwork_controller.go
pliurh: Fixed.

pkg/controller/sriovnetwork/sriovnetwork_controller.go
```diff
@@ -115,6 +116,7 @@ func (r *ReconcileSriovNetwork) Reconcile(request reconcile.Request) (reconcile.
 
 	// Fetch the SriovNetwork instance
 	instance := &sriovnetworkv1.SriovNetwork{}
+	var namespaces = ""
```
zshi-redhat: nit: namespaces -> namespace ?  This looks like will be assigned with the value of the namespace from one certain net-attach-def.

pkg/controller/sriovnetwork/sriovnetwork_controller.go
```diff
@@ -141,33 +143,59 @@ func (r *ReconcileSriovNetwork) Reconcile(request reconcile.Request) (reconcile.
...
+			reqLogger.Error(err, "Couldn't update NetworkAttachmentDefinition", "Namespace", netAttDef.Namespace, "Name", netAttDef.Name)
 			return reconcile.Result{}, err
 		}
+		// NetworkAttachmentDefinition updated successfully - don't requeue
+		return reconcile.Result{}, nil
 	}
-
-	// NetworkAttachmentDefinition already exists - don't requeue
-	reqLogger.Info("Skip reconcile: NetworkAttachmentDefinition already exists", "Namespace", found.Namespace, "Name", found.Name)
+	// networkNamespace updated, delete existing net-att-def cr and create a new one
```
zshi-redhat: There seems some code overlap that:
no matter `lastNetworkNamespace[instance.GetUID()] == netAttDef.GetNamespace()` is satisfied or not, `r.client.Get` and `r.client.Create` will be called, can we combine the two piece of code and do the conditional check only for `r.client.Delete`?

pkg/controller/sriovnetwork/sriovnetwork_controller.go
pliurh: fixed

pkg/controller/sriovnetwork/sriovnetwork_controller.go
pliurh: fixed

pkg/controller/sriovnetwork/sriovnetwork_controller.go
```diff
@@ -142,32 +144,47 @@ func (r *ReconcileSriovNetwork) Reconcile(request reconcile.Request) (reconcile.
...
 	} else {
-		reqLogger.Info("SriovNetworkNodeState already exists, updating")
-		found.Spec = netAttDef.Spec
-		err = r.client.Update(context.TODO(), found)
-		if err != nil {
-			reqLogger.Error(err, "Couldn't update SriovNetworkNodeState", "Namespace", netAttDef.Namespace, "Name", netAttDef.Name)
-			return reconcile.Result{}, err
-		}
+		lastNetworkNamespace[instance.GetUID()] = netAttDef.GetNamespace()
+		namespace = netAttDef.GetNamespace()
```
zshi-redhat: This line seems not needed.

pkg/controller/sriovnetwork/sriovnetwork_controller.go
```diff
@@ -142,32 +144,47 @@ func (r *ReconcileSriovNetwork) Reconcile(request reconcile.Request) (reconcile.
...
+			if err != nil {
+				reqLogger.Error(err, "Couldn't update NetworkAttachmentDefinition", "Namespace", netAttDef.Namespace, "Name", netAttDef.Name)
+				return reconcile.Result{}, err
+			}
+			// NetworkAttachmentDefinition updated successfully - don't requeue
+			return reconcile.Result{}, nil
+		} else {
+			err = r.client.Delete(context.TODO(), &netattdefv1.NetworkAttachmentDefinition{
+			    ObjectMeta: metav1.ObjectMeta{
+					Name: netAttDef.GetName(),
```
zshi-redhat: nit: there is extra space at the end of this line (`Name: netAttDef.GetName(),`).

pkg/controller/sriovnetwork/sriovnetwork_controller.go
pliurh: fixed.

pkg/controller/sriovnetwork/sriovnetwork_controller.go
pliurh: fixed

PR #54, opened 2019-08-05
Don't crash if grubby is not installed

fedepaol:
The generic plugin uses grubby to check if `intel_iommu=on` and `iommu=pt` are set as kernel parameters.

This assumes that grubby is installed and that the current host is rhel based, which might not be the case in upstream. 
This pr checks for grubby existance and if it's not installed simply raises an error without having the node config daemon restart

### DIFF
--- a/bindata/scripts/enable-kargs.sh
+++ b/bindata/scripts/enable-kargs.sh
@@ -5,8 +5,7 @@ chroot /host/ which grubby
 
 # if grubby is not there, let's send a message
 if [ $? -ne 0 ]; then
-    echo "grubby not available"
-    exit
+    exit 127
 fi
 
 declare -a kargs=( "$@" )
--- a/bindata/scripts/enable-kargs.sh
+++ b/bindata/scripts/enable-kargs.sh
@@ -14,7 +14,7 @@ if grep --quiet CoreOS "$REDHAT_RELEASE_FILE"; then
         fi
     done
 else
-    chroot /host/ which grubby
+    chroot /host/ which grubby > /dev/null 2>&1
     # if grubby is not there, let's tell it
     if [ $? -ne 0 ]; then
         exit 127
--- a/bindata/scripts/enable-kargs.sh
+++ b/bindata/scripts/enable-kargs.sh
@@ -1,5 +1,13 @@
 #!/bin/bash
 set -x
+
+chroot /host/ which grubby
+
+# if grubby is not there, let's send a message
+if [ $? -ne 0 ]; then
+    echo "grubby not available"
+fi
+
 declare -a kargs=( "$@" )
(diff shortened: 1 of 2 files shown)


pkg/plugins/generic/generic_plugin.go
```diff
@@ -20,7 +20,10 @@ type GenericPlugin struct {
 	LoadVfioDriver uint
 }
 
-const scriptsPath = "bindata/scripts/enable-kargs.sh"
+const (
+	scriptsPath          = "bindata/scripts/enable-kargs.sh"
+	commandNotFoundError = 127
```
zshi-redhat: Where this const is used?

zshi-redhat:
> In which case grubby is not installed?

ok, this is for upstream which may not use rhel based distro I guess

fedepaol:
> > In which case grubby is not installed?
> 
> ok, this is for upstream which may not use rhel based distro I guess

Exactly. For example, kind nodes images (which I am using for CI) are debian based.

pkg/plugins/generic/generic_plugin.go
fedepaol: Nowhere anymore. Was using it for checking the error code but realized the daemon image is being built with go 1.10 which does not yet have `ExitCode()` method on  `ExitError`s.
Will remove it.

bindata/scripts/enable-kargs.sh
```diff
@@ -12,6 +14,12 @@ if grep --quiet CoreOS "$REDHAT_RELEASE_FILE"; then
         fi
     done
 else
+    chroot /host/ which grubby
```
zshi-redhat: Shall we redirect the stdout and stderr for the `which` cmd? otherwise it may appear in the exit code and break the check condition.

fedepaol: Oh, you are right. I just tested (my) happy path where which was failing.

pkg/plugins/generic/generic_plugin.go
```diff
@@ -170,6 +170,12 @@ func tryEnableIommuInKernelArgs() (bool, error) {
 	cmd.Stderr = &stderr
 
 	if err := cmd.Run(); err != nil {
+		// if grubby is not there log and assume kernel args are set correctly.
+		if exiterr, ok := err.(*exec.ExitError); ok && exiterr.Error() == "exit status 127" {
```
pliurh: I think it would be better to check the exit status with
```
        if status, ok := exitErr.Sys().(syscall.WaitStatus); ok &&  status.ExitStatus() == 127 {
            ...
        }
```

fedepaol: Thanks!

PR #55, opened 2019-08-06
Fix kargs loading issue on RHCOS

pliurh:
Fix BZ1734626


zshi-redhat:
Is this fix related to BZ1734626?

pliurh:
Yes. #46 is not enough to fix that BZ.

PR #56, opened 2019-08-06
Update entire DaemonSet CR

pliurh:
Fix https://bugzilla.redhat.com/show_bug.cgi?id=1737780

### DIFF
--- a/pkg/controller/sriovnetworknodepolicy/sriovnetworknodepolicy_controller.go
+++ b/pkg/controller/sriovnetworknodepolicy/sriovnetworknodepolicy_controller.go
@@ -19,7 +19,6 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
-	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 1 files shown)


pkg/controller/sriovnetworknodepolicy/sriovnetworknodepolicy_controller.go
```diff
@@ -755,10 +755,8 @@ func (r *ReconcileSriovNetworkNodePolicy) syncDaemonSet(cr *sriovnetworkv1.Sriov
 		}
 	} else {
 		logger.Info("DaemonSet already exists, updating")
-		if ds.Spec.Template.Spec.Affinity != nil && in.Spec.Template.Spec.Affinity != nil {
-			if !reflect.DeepEqual(*ds.Spec.Template.Spec.Affinity.NodeAffinity, *in.Spec.Template.Spec.Affinity.NodeAffinity) {
-				ds.Spec.Template.Spec.Affinity.NodeAffinity = in.Spec.Template.Spec.Affinity.NodeAffinity
-			}
+		if !apiequality.Semantic.DeepEqual(ds.Spec, in.Spec) {
```
zshi-redhat: Why need to compare the difference, can we just do the `client.Update` and let the api to figure out whether it should do the real update or not?

pliurh: Good point. Fixed.

PR #58, opened 2019-08-08
Bug 1738421: Re-create default nodePolicy CR when it deleted

pliurh:
Also let operator watch secondary resources, so that secondary resondary resource can be restored.

Fix https://bugzilla.redhat.com/show_bug.cgi?id=1738421


pkg/controller/sriovnetworknodepolicy/sriovnetworknodepolicy_controller.go
```diff
@@ -106,6 +176,22 @@ func (r *ReconcileSriovNetworkNodePolicy) Reconcile(request reconcile.Request) (
 	defaultPolicy := &sriovnetworkv1.SriovNetworkNodePolicy{}
 	err := r.client.Get(context.TODO(), types.NamespacedName{Name: DEFAULT_POLICY_NAME, Namespace: Namespace}, defaultPolicy)
 	if err != nil {
+		if errors.IsNotFound(err) {
+			// Default policy object not found, create it.
+			defaultPolicy.SetNamespace(Namespace)
```
zshi-redhat: Is it possible that namespace from reconcile.request is not from `sriov-network-operator`?

pliurh: Currently, the operator is designed to only work in namespace `sriov-network-operator`.

pliurh:
@zshi-redhat fixed

PR #63, opened 2019-08-10
wip(x520): adding support for X520 dual-10G DA/SFP+

faust64:
Purely theoretical, although based on successful tests on 4.1, involving openshift/sriov-network-device-plugin#16

### DIFF
--- a/pkg/apis/sriovnetwork/v1/helper.go
+++ b/pkg/apis/sriovnetwork/v1/helper.go
@@ -7,7 +7,7 @@ import (
 
 var SriovPfVfMap = map[string](string){
 	"1583": "154c",
-	"154d": "10ed",
+	"154b": "10ed",
 	"10fb": "10ed",
 	"1015": "1016",
 	"1017": "1018",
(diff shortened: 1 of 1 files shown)


fepan:
There is no current plan to support X520.

pliurh:
@fepan shall we be open to this in upstream if there're people want to contribute? We can still claim that we don't support it downstream.

> There is no current plan to support X520.

faust64:
Indeed, refusing contributions to an opensource project because "there's no plan to support X" doesn't make much sense. Or at least, goes against the "upstream first" approach.

I'm sure the average OKD user don't care much about which devices would be supported, should they go with OCP.
While the only OCP 4 customer I've met so far was wondering: why would we need card-by-card approval?

I'm sure there would be some discrepancies from vendor to vendor, still I doubt the X520 chip from Intel varies much from that X710 one we're supporting already.
IMHO, considering most OCP 4.x customers are only evaluating OpenShift 4, it might be better not to limit them right away, rather encourage them in giving us some feedback for cases that were not otherwise validated. Which might even help in cherry-picking whichever new devices to support in future releases.

fepan:
The whitelist behavior is only implemented downstream for support reasons. There is no restriction on upstream as to which NIC is used.

eparis:
at this moment PRs should be sent to master, not to release-4.2. I believe this should be closed...

fepan:
We have no plan to implement this feature at this time. Closing.

PR #66, opened 2019-08-20
Add test cases for update/delete SriovNetwork CR

pliurh:


zshi-redhat:
Are these test supposed to run locally or in a real environment?

pliurh:
> Are these test supposed to run locally or in a real environment?

The answer is both. They're tests for the API.

Issue #51, opened 2019-08-02
No synchronization between nodes in case of reboot


fedepaol:
In case the change of the NetworkNodePolicy trigger a reboot of the node (i.e. to add the needed kernel parameters, or to re-set a mellanox card), the reboot of the node is immediate.

This risks to jeopardize a zone if the same policy is applied at the same time to all the nodes. 
The reboot should be gradual and involve a small group of nodes at the same time, while keeping the zone functioning.

pliurh:
I haven't test this scenario in a large scale cluster myself. Here is the design logic.

Before rebooting, node would be drained. Only if all pods get evicted successfully, node can reboot. If there's no available capacity in cluster, pod eviction will fail. So node shall not reboot all at the same time.

fedepaol:
I saw the drain happening before, but I don't see checking if it fails and avoiding the reboot (but I might be missing something).

The code should be around here:
https://github.com/openshift/sriov-network-operator/blob/212dfea945e8ebe857d3b14d44edd43d4233093f/pkg/daemon/daemon.go#L265
but no error is returned there.

pliurh:
There is a back-off mechanism in function drainNode(). It will keep on retry, until node is drained. Currently this function is implemented by MCO team. And, there is a proposal in upstream https://github.com/kubernetes/enhancements/pull/1080. We will switch to the native API, when it gets ready.

fedepaol:
I saw that but I did not notice how long was the waiting time.
Guess this answers my question, thanks for the explanation.

PR #67, opened 2019-08-21
Add a way to disable node reboot

fedepaol:
We are planning to use the operator to set up sriov in kubevirt CI environment.
Since the whole setup is a kind cluster spun up as a prow job, a reboot of what the operator thinks is a node would be a reboot of the whole CI node.

With this PR, if a configmap is present, the node configuration daemon will log an error and exit instead of rebooting the node, so that the user of CI can check what is missing and schedule an intervention.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -18,10 +18,12 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
+
 	// "k8s.io/client-go/informers"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/cache"
+
 	// "k8s.io/client-go/kubernetes/scheme"
 	sriovnetworkv1 "github.com/openshift/sriov-network-operator/pkg/apis/sriovnetwork/v1"
 	snclientset "github.com/openshift/sriov-network-operator/pkg/client/clientset/versioned"
@@ -55,9 +57,9 @@ type Daemon struct {
 
 	refreshCh chan<- Message
 
-	dpReboot bool
-
 	mu *sync.Mutex
+
+	noReboot bool
 }
 
 const scriptsPath = "/bindata/scripts/enable-rdma.sh"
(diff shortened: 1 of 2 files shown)


pkg/daemon/daemon.go
```diff
@@ -381,7 +395,7 @@ func rebootNode() {
 	// as systemd will time out the stop invocation.
 	cmd := exec.Command("systemd-run", "--unit", "sriov-network-config-daemon-reboot",
```
booxter: would it make sense to expose this command through configuration? In this way not only you would be able to disable node reboot completely, but others could adjust how the reboot is achieved. (Perhaps because the platform doesn't use systemd, or some additional steps should be executed before calling to 'reboot'.)

Of course, this is up to maintainers of the project to decide how far they want to go.

fedepaol: From my point of view it would make sense, but I would treat it as a separate pr / issue.

pliurh:
I am not quite sure if we really need the ability to disable reboot. Currently, the config daemon will only trigger reboot in following cases:
1. For Mellanox NIC, the desired VF number is larger than what has totalVFs.
2. Updating kargs to enable IOMMU on host for vfio driver. 

In those cases, reboot is mandatory, otherwise, the config would not take effect.

booxter:
@pliurh even if a reboot is ultimately needed, it's unsafe to trigger reboots for all nodes at the same time and the admin may want to orchestrate the reboots properly. (I also think it's not "polite" for an operator to reboot a node without an opt-out but that's more subjective.)

fedepaol:
@booxter, @pliurh explained the rationale behind the rebooting in https://github.com/openshift/sriov-network-operator/issues/51 , and he managed to convince me :smile:

booxter:
@fedepaol one (other) issue with draining approach is that it makes it impossible to run the operator on a single node cluster because there are no places to drain resources to, or in clusters where master is also a worker (because it's impossible to drain api server etc. from the node).

fedepaol:
@pliurh , any chances to get this merged in the next iteration?

booxter:
@pliurh how can we help with getting this merged? Or is there some release-related blocker that I am not aware of that has to postpone consideration for this PR?

pliurh:
@fedepaol Could you expose this flag in the [new operator config API]( https://github.com/openshift/sriov-network-operator/pull/94) instead, I think it is better we manage all the flags in one place. Thanks.

fedepaol:
> @fedepaol Could you expose this flag in the [new operator config API](https://github.com/openshift/sriov-network-operator/pull/94) instead, I think it is better we manage all the flags in one place. Thanks.

Will do, thanks for the feedback

PR #70, opened 2019-08-28
BUG 1746222: Change default namespace to openshift-sriov-network-operator

pliurh:


zshi-redhat:
> /hold

why hold? Is this about related to adding metrics in sriov operator?

pliurh:
> > /hold
> 
> why hold? Is this about related to adding metrics in sriov operator?

Yes. In one hand, the monitoring team told me that creating a `openshift-*` namespace by user is not recommend. In other hand,  cluster monitoring can only work with `openshift-*` namespace components. So we need to wait for their decision on this.

zshi-redhat:
By changing the namespace, how do we consider upgrade SR-IOV Operator from 4.2 to 4.3 version?

pliurh:
> By changing the namespace, how do we consider upgrade SR-IOV Operator from 4.2 to 4.3 version?

It is impossible to upgrade with namespace change. Users have to re-install instead. However, since 4.2 is TP for SRIOV, I think it is acceptable.

zshi-redhat:
so the differs from previous change is just re-adding `deploy/crds/sriovnetwork_v1_sriovnetwork_cr.yaml` file?

pliurh:
> so the differs from previous change is just re-adding `deploy/crds/sriovnetwork_v1_sriovnetwork_cr.yaml` file?

It is an error during rebase. This file has been replaced by "sriovnetwork.openshift.io_v1_sriovnetwork_cr.yaml", during the operator-sdk update.

PR #72, opened 2019-08-29
BUG 1746329: Not create NetworkAttachmentDefinition CR if already exists

pliurh:


pkg/controller/sriovnetwork/sriovnetwork_controller.go
```diff
@@ -148,6 +148,18 @@ func (r *ReconcileSriovNetwork) Reconcile(request reconcile.Request) (reconcile.
...
 	}
+	
+	// Try initialize lastNetworkNamespace
+	nadList := &netattdefv1.NetworkAttachmentDefinitionList{}
+	fieldSelector := fields.Set{"metadata.name": instance.GetName(),}.AsSelector()
+	r.client.List(context.TODO(), &client.ListOptions{FieldSelector: fieldSelector,}, nadList)
+	reqLogger.Info("NetworkAttachmentDefinition", "list", nadList)
+	if len(nadList.Items) == 1 {
+		lastNetworkNamespace[instance.GetUID()] = nadList.Items[0].GetNamespace()
+	} else if len(nadList.Items) > 1 {
```
zshi-redhat: Will we allow multiple net-attach-def with same name in different namespaces?

pliurh: No. Since the net-att-def will use the same name as the SriovNetwork CR, and all the SriovNetwork CRs are in one namespace.

zshi-redhat: so all the custom resources of SR-IOV net-attach-def will be retricted in one namespace?

pliurh: No, all the SriovNetwork CRs will be in one namespace. You can create net-att-def CR in arbitrary namespace by specifying the `networkNamespace`of the SriovNetwork CR.

zshi-redhat: ok, so in above code, when `r.client.List` lists all the net-attach-def instances, it may contain two net-attach-def instances with same name, but different namespaces. Then will the condition `len(nadList.Items) > 1` be evaluated as true ?

pliurh: Yes, it is possible, if the net-att-def CR is created by other components or human. In such case, the operator will throw error. User shall remove the naming conflict manually.

zshi-redhat: net-attach-net will have the same name as `sriovnetwork` custom resource, since `sriovnetwork` can only be created in one namespace, so we cannot have net-attach-net with same names in different namespaces. but net-attach-def can be created in any namespace.

PR #73, opened 2019-09-01
Support change link type to ETH or IB

Mmduh-483:


pkg/plugins/mellanox/mellanox_plugin.go
```diff
@@ -99,9 +98,19 @@ func (p *MellanoxPlugin) OnNodeStateChange(old, new *sriovnetworkv1.SriovNetwork
...
 			requireChange = true
 		}
-		if fwCurrent.linkType != EthLinkType && fwCurrent.linkType != PreconfiguredLinkType && fwCurrent.linkType != UknownLinkType {
-			attrs.linkType = EthLinkType
+		if iface.LinkType != EthLinkType && iface.LinkType != InfinibandLinkType {
+			err = fmt.Errorf("mellanox-plugin OnNodeStateChange(): Not supported link type: %s", iface.LinkType)
+			return false, false, err
+		}
+		if fwCurrent.linkType == PreconfiguredLinkType || fwCurrent.linkType == UknownLinkType {
+			err = fmt.Errorf("mellanox-plugin OnNodeStateChange(): Preconfigured NIC or Unkown link type: %s", fwCurrent.linkType)
```
pliurh: What's the meaning of `Preconfigured NIC`? The error message looks a bit confusing, why a `Preconfigured NIC` would lead  to an error?

deploy/crds/sriovnetwork_v1_sriovnetworknodepolicy_crd.yaml
```diff
@@ -39,6 +39,9 @@ spec:
             isRdma:
               description: RDMA mode. Defaults to false.
               type: boolean
+            linkType:
+              description: Link Type for NIC
```
pliurh: It would be nice put the allowed value in the description. Also an `enum` of allowed values can help the API validation.

pkg/plugins/mellanox/mellanox_plugin.go
Mmduh-483: In some of mlnx's cards the `LINK_TYPE_P*` is reconfigured to IB or ETH and can't be changed

deploy/crds/sriovnetwork_v1_sriovnetworknodepolicy_crd.yaml
Mmduh-483: Good idea

Mmduh-483:
@pliurh can you please comment on this patch

pkg/plugins/mellanox/mellanox_plugin.go
adrianchiris: I would separate the two cases

in case of unknown link type:
`err = fmt.Errorf("mellanox-plugin OnNodeStateChange(): Unknown Link type : %s", fwCurrent.linkType)`

in case of a PreConfigredLinkType:
`err = fmt.Errorf("mellanox-plugin OnNodeStateChange(): Network card does not support Link type change.")`

pkg/plugins/mellanox/mellanox_plugin.go
```diff
@@ -99,9 +98,21 @@ func (p *MellanoxPlugin) OnNodeStateChange(old, new *sriovnetworkv1.SriovNetwork
 			attrs.enableSriov = true
 			requireChange = true
 		}
-		if fwCurrent.linkType != EthLinkType && fwCurrent.linkType != PreconfiguredLinkType && fwCurrent.linkType != UknownLinkType {
-			attrs.linkType = EthLinkType
-			requireChange = true
+		if iface.LinkType != "" {
```
adrianchiris: So you defined a new field in the CRD, 
what is the expected behaviour if this field is not set ?

previously mellanox plugin configured the link type to Ethernet.
now it will simply do nothing. 
This change is not backward compatible right ? (which is something i think we want)

adrianchiris: Also, a separate method handling this logic will help with clarity.

pkg/plugins/mellanox/mellanox_plugin.go
```diff
@@ -99,9 +98,21 @@ func (p *MellanoxPlugin) OnNodeStateChange(old, new *sriovnetworkv1.SriovNetwork
...
+			if iface.LinkType != EthLinkType && iface.LinkType != InfinibandLinkType {
+				err = fmt.Errorf("mellanox-plugin OnNodeStateChange(): Not supported link type: %s", iface.LinkType)
+				return false, false, err
+			}
+			if fwCurrent.linkType == PreconfiguredLinkType || fwCurrent.linkType == UknownLinkType {
+				err = fmt.Errorf("mellanox-plugin OnNodeStateChange(): Preconfigured NIC or Unkown link type: %s", fwCurrent.linkType)
```
adrianchiris: I would separate the error in case LinkType is not configurable and Unknown link type.

also a better name would be `FixedLinkType` or `StaticLinkType`. the name `PreConfigured` does not hint in any way that the configuration cannot be changed.

pkg/utils/utils.go
```diff
@@ -82,6 +83,17 @@ func DiscoverSriovDevices() ([]sriovnetworkv1.InterfaceExt, error) {
 			iface.Name = name
 		}
 
+		if iface.Name != "" {
+			if devLink, err := netlink.LinkByName(iface.Name); err == nil {
```
adrianchiris: devlink ?  maybe just `link` ? as to not confuse with linux devlink subsystem.
netlink.LinkByName retrieves rtnl link device

pkg/utils/utils.go
```diff
@@ -118,6 +130,16 @@ func SyncNodeState(newState *sriovnetworkv1.SriovNetworkNodeState) error {
 					glog.Errorf("SyncNodeState(): fail to config sriov interface %s: %v", iface.PciAddress, err)
 					return err
 				}
+				if ifaceStatus.Name != "" {
+					if devLink, err := netlink.LinkByName(ifaceStatus.Name); err == nil {
```
adrianchiris: same as above.

also a helper method of : getLinkType() would reduce code duplication

pkg/plugins/mellanox/mellanox_plugin.go
Mmduh-483: I will change it to StaticLinkType

pkg/plugins/mellanox/mellanox_plugin.go
Mmduh-483: The behavior will be if there is no link type specified the plugin will assume it to be ETH

Mmduh-483:
@adrianchiris All previous comments are fixed now
@pliurh Do you have any comments?

deploy/crds/sriovnetwork_v1_sriovnetworknodestate_crd.yaml
```diff
@@ -51,10 +53,6 @@ spec:
           type: object
         status:
           properties:
-            lastSyncError:
-              type: string
-            syncStatus:
-              type: string
```
Mmduh-483: @pliurh each time I run `make gencode` there are these changes, is right? or there is something needed to be changed?

pkg/plugins/mellanox/mellanox_plugin.go
adrianchiris: > Also, a separate method handling this logic will help with clarity.

I still think this is desirable, but will not insist on it...

pkg/plugins/mellanox/mellanox_plugin.go
```diff
@@ -99,9 +98,32 @@ func (p *MellanoxPlugin) OnNodeStateChange(old, new *sriovnetworkv1.SriovNetwork
...
+				err = fmt.Errorf("mellanox-plugin OnNodeStateChange(): Network card %s does not support link type chage", iface.PciAddress)
+				return false, false, err
+			}
+			if fwNext.linkType != iface.LinkType {
+				attrs.linkType = iface.LinkType
+				requireChange = true
+			} else if fwCurrent.linkType != fwNext.linkType {
+				needReboot = true
+			}
+		} else if fwCurrent.linkType != UknownLinkType && fwCurrent.linkType != StaticLinkType {
```
adrianchiris: for clarity i would:

```
if iface.LinkType != "" {
    ...
    ...
} else {
    // If linkType was not specified in CRD, set link type to Ethernet if needed
    if fwCurrent.linkType == UnknownLinkType || fwCurrent.linkType == StaticLinkType {
        // do nothing
        fmt.Infof("Link type for NIC is assumed to be Ethernet.")
    } else  {
        if fwNext.linkType != EthLinkType {
            attrs.linkType = EthLinkType
            requireChange = true
        } else if fwCurrent.linkType != EthLinkType {
        needReboot = true
    }
}
```

pkg/plugins/mellanox/mellanox_plugin.go
```diff
@@ -25,15 +25,15 @@ type mlnxNic struct {
...
-	InfinibandLinkType    = "IB"
-	PreconfiguredLinkType = "Preconfigured"
-	UknownLinkType        = "Uknown"
-	TotalVfs              = "NUM_OF_VFS"
-	EnableSriov           = "SRIOV_EN"
-	LinkTypeP1            = "LINK_TYPE_P1"
-	LinkTypeP2            = "LINK_TYPE_P2"
-	MellanoxVendorId      = "15b3"
+	EthLinkType        = "ETH"
+	InfinibandLinkType = "IB"
```
adrianchiris: maybe move EthLinkType and InfinibandLinkType to utils/consts.go or similar
and use them in both mellanox plugin and utils.go (getLinkType())

Mmduh-483: It was meant for mellanox plugin to not depend on the utils, the `getlinkType()` in utils is for Discovery and update the status for the NICs of the Node

Mmduh-483:
@moshe010

deploy/crds/sriovnetwork.openshift.io_sriovnetworknodepolicies_crd.yaml
```diff
@@ -43,6 +43,12 @@ spec:
             isRdma:
               description: RDMA mode. Defaults to false.
               type: boolean
+            linkType:
+              description: Link Type for NIC. Allowed value "ETH", "IB".
```
moshe010, line 47: s/Link Type for NIC/NIC Link Type

deploy/crds/sriovnetwork.openshift.io_v1_sriovnetworknodepolicy_cr.yaml
```diff
@@ -32,3 +32,22 @@ spec:
     rootDevices: ['0000:02:00.0', '0000:02:00.1']
   deviceType: netdevice
   isRdma: true
+
+---
+# Mlnx IB Nics doesn't support changing mtu
```
moshe010, line 37: it supported but it is diffrent from ethernet and requires configuration in opensm. Currently we can say it out of scope

pkg/apis/sriovnetwork/v1/zz_generated.openapi.go
```diff
@@ -184,6 +184,13 @@ func schema_pkg_apis_sriovnetwork_v1_SriovNetworkNodePolicySpec(ref common.Refer
 							Format:      "",
 						},
 					},
+					"linkType": {
+						SchemaProps: spec.SchemaProps{
+							Description: "Link Type for NIC. Allowed value \"ETH\", \"IB\".",
```
moshe010, line 189: Please update Description as suggested above

pkg/plugins/mellanox/mellanox_plugin.go
```diff
@@ -76,32 +76,41 @@ func (p *MellanoxPlugin) OnNodeStateChange(old, new *sriovnetworkv1.SriovNetwork
...
-		} else if iface.NumVfs > 0 && !fwCurrent.enableSriov {
+		} else if iface.NumVfs > 0 && !fwData.enableSriov {
 			attrs.enableSriov = true
 			requireChange = true
 		}
-		if fwCurrent.linkType != EthLinkType && fwCurrent.linkType != PreconfiguredLinkType && fwCurrent.linkType != UknownLinkType {
-			attrs.linkType = EthLinkType
-			requireChange = true
+		if iface.LinkType != "" {
+			if iface.LinkType != EthLinkType && iface.LinkType != InfinibandLinkType {
```
moshe010, line 98: Do you need to create this PR on top of the dual port fix you have?

deploy/crds/sriovnetwork.openshift.io_v1_sriovnetworknodepolicy_cr.yaml
```diff
@@ -32,3 +32,22 @@ spec:
...
+apiVersion: sriovnetwork.openshift.io/v1
+kind: SriovNetworkNodePolicy
+metadata:
+  name: mlnx-ib-nics
+  namespace: sriov-network-operator
+spec:
```
pliurh, line 43: `linkType` shall be specified in this CR, right?

Mmduh-483: yes, added

Mmduh-483:
> I don't see the controller side change. Do you plan to put it in another PR?

What changes?

pkg/plugins/mellanox/mellanox_plugin.go
Mmduh-483: Yes, It can be done after one of these 2 get merged since there will be a conflict between them

Mmduh-483: also it will be the same case for the switchdev mode PR

Mmduh-483:
@pliurh @zshi-redhat Can you please give a review

zshi-redhat:
> > I don't see the controller side change. Do you plan to put it in another PR?
> 
> What changes?

The config data of sriov device plugin needs to be generated with LinkType selector. The logic is contained in node policy controller. (It can be added in a separate PR and focus this one for MLX plugin change)

deploy/crds/sriovnetwork.openshift.io_v1_sriovnetworknodepolicy_cr.yaml
```diff
@@ -32,3 +32,23 @@ spec:
     rootDevices: ['0000:02:00.0', '0000:02:00.1']
   deviceType: netdevice
   isRdma: true
+
+---
+# Changing mtu for Mellanox IB NICs is out of scope
```
zshi-redhat, line 37: Does the annotation mean when IB linkType is specified, mtu should not be configured to a non-default value?
Why is it out of scope?

deploy/crds/sriovnetwork.openshift.io_sriovnetworknodepolicies_crd.yaml
```diff
@@ -43,6 +43,12 @@ spec:
             isRdma:
               description: RDMA mode. Defaults to false.
               type: boolean
+            linkType:
+              description: NIC Link Type. Allowed value "ETH", "IB".
+              enum:
+                - ETH
```
zshi-redhat, line 49: Shall we also allow lower case and full names such as `eth`, `ether`, `ib`, `infiniband` etc?

zshi-redhat:
Note: we should also do validation of IB config in separate PR, such as mtu is in the supported range, supported device ID for IB device etc.

pkg/plugins/mellanox/mellanox_plugin.go
```diff
@@ -25,15 +25,15 @@ type mlnxNic struct {
 }
 
 const (
-	EthLinkType           = "ETH"
-	InfinibandLinkType    = "IB"
-	PreconfiguredLinkType = "Preconfigured"
```
zshi-redhat, line 30: Is this replaced by `StaticLinkType`? or does it only applied to certain kind of devices?

pkg/plugins/mellanox/mellanox_plugin.go
```diff
@@ -76,32 +76,41 @@ func (p *MellanoxPlugin) OnNodeStateChange(old, new *sriovnetworkv1.SriovNetwork
 		if !isMlnxNicAndInNode(iface.PciAddress, new) {
 			continue
 		}
-		fwCurrent, fwNext, err := getMlnxNicFwData(iface.PciAddress)
+		fwData, err := getMlnxNicFwData(iface.PciAddress)
 		if err != nil {
 			return false, false, err
 		}
-		if fwCurrent.enableSriov != fwNext.enableSriov || fwCurrent.totalVfs != fwNext.totalVfs ||
```
zshi-redhat, line 83: Why this is not needed? Does it compare totalVfs between Current and Next somewhere?

deploy/crds/sriovnetwork.openshift.io_v1_sriovnetworknodepolicy_cr.yaml
moshe010: The MTU is configured by the SM. A user can put a value which is less or equal to the MTU configured in the SM if he want.

deploy/crds/sriovnetwork.openshift.io_sriovnetworknodepolicies_crd.yaml
moshe010: maybe just ib and eth so we will suppot (IB/ETH/ib/eth)

pkg/plugins/mellanox/mellanox_plugin.go
moshe010: So I think it better that we first merge the PR with the dual port fix and then @Mmduh-483 wll rebase this PR.

Mmduh-483:
> LinkType

It will be in a another PR

pkg/plugins/mellanox/mellanox_plugin.go
Mmduh-483: It is just replacing of the variable, not related to devices

deploy/crds/sriovnetwork.openshift.io_v1_sriovnetworknodepolicy_cr.yaml
pliurh: @moshe010 Do you mean the SM will configure the MTU of the VFs/PFs automatically, so the operator doesn't need to set it?

deploy/crds/sriovnetwork.openshift.io_sriovnetworknodepolicies_crd.yaml
```diff
@@ -37,12 +37,20 @@ spec:
               description: The driver type for configured VFs. Allowed value "netdevice",
                 "vfio-pci". Defaults to netdevice.
               enum:
-              - netdevice
-              - vfio-pci
+                - netdevice
```
pliurh, line 40: Did you use operator-sdk to generate the CRDs? Why adding space here?

pkg/plugins/mellanox/mellanox_plugin.go
Mmduh-483: @zshi-redhat this check won't be needed anymore, there is 2 checks (link type removed in this PR) and the (sriovTotal and sriovEnable) will be removed in the reset the FW total VFs

deploy/crds/sriovnetwork.openshift.io_sriovnetworknodepolicies_crd.yaml
Mmduh-483: Yes, it is a space generate by auto format tool I use, will remove it

Mmduh-483: I see it is adding some fields and removing other fields also it is changing the `apiVersion`, can you please cooperate, which changes to keep, attached the diff file when run `make gencode'

pkg/apis/sriovnetwork/v1/sriovnetworknodepolicy_types.go
```diff
@@ -33,6 +33,9 @@ type SriovNetworkNodePolicySpec struct {
 	DeviceType string `json:"deviceType,omitempty"`
 	// RDMA mode. Defaults to false.
 	IsRdma bool `json:"isRdma,omitempty"`
+	// +kubebuilder:validation:Enum=eth;ETH;ib;IB
+	// NIC Link Type. Allowed value "eth", "ETH", "ib", and "IB".
```
adrianchiris, line 37: I think you should swap the comments so that CRD generation adds the enum to the schema

Mmduh-483: It is adding the enum like that

deploy/crds/sriovnetwork.openshift.io_sriovnetworknodepolicies_crd.yaml
adrianchiris: I get the feeling code generation might change between operator-sdk versions....

@pliurh does this project have any guideline on how to update CRDs after changing the corresponding `_types.go` file

pkg/apis/sriovnetwork/v1/sriovnetworknodepolicy_types.go
adrianchiris: missed that :) NVM.

deploy/crds/sriovnetwork.openshift.io_sriovnetworknodepolicies_crd.yaml
pliurh: That is true. Here are some guideline I can think of:

1. Normally we shall use the `operator-sdk` command with the same version as https://github.com/openshift/sriov-network-operator/blob/97d4e33c781224c0758f15c05619fa78f2c5567b/go.mod#L21
2. the NetworkAttachmentDefinition shall always be ignored, as it was not generated by `operator-sdk`. 
3. `preserveUnknownFields: false` shall be kept. Otherwise, `oc explain` wouldn't work.
4. Don't touch any existing fields if it is not necessary.

Mmduh-483:
@pliurh @zshi-redhat Can you please re-review the PR

zshi-redhat:
The change of linkType API looks good to me. 
/lgtm

Several questions that may need to address with follow-up patch:
1) Do we need a validation in generic plugin for IB mtu setting (I understand we can do it in webhook, but webhook can be disabled)?
2) In which case will the name IB device be changed after rebooting? Does it matter if ib device name is not consistent across rebooting?

PR #75, opened 2019-09-08
Support changing sriov mode to legacy or switchdev for ETH devices

Mmduh-483:


deploy/crds/sriovnetwork_v1_sriovnetworknodepolicy_crd.yaml
```diff
@@ -36,6 +36,9 @@ spec:
               - netdevice
               - vfio-pci
               type: string
+            enableSwitchDev:
+              description: Enable Switch Dev mode for Mellnaox's cards
```
moshe010: s/ Switch Dev/ Switchdev. It is not specific to mellanox ...

moshe010: It better to do enum and called it SR-IOV mode (legacy/switchdev) the default should be legacy

pkg/plugins/mellanox/mellanox_plugin.go
```diff
@@ -273,3 +280,59 @@ func isMlnxNicAndInNode(pciAddress string, state *sriovnetworkv1.SriovNetworkNod
 	}
 	return false
 }
+
+func isSwitchDevModeEnabled(pciAddress string) (bool, error) {
+	glog.Infof("mellanox-plugin isSwitchDevModeEnabled(): device %s", pciAddress)
+	args := []string{"dev", "eswitch", "show", fmt.Sprintf("pci/%s", pciAddress)}
```
moshe010: please move this from the Mellanox to generic plugin. also please use netlink to get/set switchdev mode.
https://github.com/vishvananda/netlink/commit/fd97bf4e47867b5e794234baa6b8a7746135ec10

pkg/utils/utils.go
```diff
@@ -346,3 +361,36 @@ func Chroot(path string) (func() error, error) {
...
+
+	err := cmd.Run()
+	glog.V(2).Infof("RunCommand(): %s", stdout.String())
+	return stdout.String(), err
+}
+
+func GetNicSriovMode(pciAddress string) (string, error) {
+	glog.V(2).Infof("GetNicSriovMode(): device %s", pciAddress)
+	args := []string{"dev", "eswitch", "show", fmt.Sprintf("pci/%s", pciAddress)}
+	out, err := RunCommand("devlink", args...)
```
moshe010: Why not using the go netlink lib DevLinkSetEswitchMode and DevLinkGetDeviceByName see https://github.com/vishvananda/netlink/blob/fd97bf4e47867b5e794234baa6b8a7746135ec10/devlink_linux.go

pkg/utils/utils.go
```diff
@@ -346,3 +361,36 @@ func Chroot(path string) (func() error, error) {
...
+	if err != nil {
+		return "", err
+	}
+	if strings.Contains(out, "switchdev") {
+		return "switchdev", nil
+	}
+	return "legacy", nil
+}
+
+func ChangeNicSriovMode(pciAddress, mode string) error {
```
moshe010: s/ChangeNicSriovMode/setNicSriovMode/

pkg/utils/utils.go
```diff
@@ -169,6 +176,14 @@ func configSriovDevice(iface *sriovnetworkv1.Interface, ifaceStatus *sriovnetwor
 		if err != nil {
 			glog.Warningf("configSriovDevice(): unable to parse VFs for device %+v %q", iface.PciAddress, err)
 		}
+		for _, vf := range vfs {
+			if err := Unbind(vf); err != nil{
+				glog.Warningf("configSriovDevice(): fail to unbind driver for device %s", vf)
+			}
+		}
+		if err := ChangeNicSriovMode(iface.PciAddress, iface.SriovMode); err != nil {
```
moshe010: After moving to switchdev mode you need to set ethtool -K <pf interface> hw-tc-offload on

Mmduh-483:
@pliurh Can you please comment on this patch?

pkg/utils/utils.go
```diff
@@ -169,6 +177,24 @@ func configSriovDevice(iface *sriovnetworkv1.Interface, ifaceStatus *sriovnetwor
...
+					}
+				}
+				if err := setNicSriovMode(iface.PciAddress, iface.SriovMode); err != nil {
+					return err
+				}
+				sriovMode, err := GetNicSriovMode(iface.PciAddress)
+				if err != nil {
+					return err
+				}
+				ifaceStatus.SriovMode = sriovMode
```
pliurh: The content of Status shall be update in DiscoverSriovDevices(), this function shall only configure things.

pkg/utils/utils.go
```diff
@@ -169,6 +177,24 @@ func configSriovDevice(iface *sriovnetworkv1.Interface, ifaceStatus *sriovnetwor
...
 		}
+		// Set Sriov mode for ETH devices
+		if ifaceStatus.Name != "" {
+			if link, err := netlink.LinkByName(ifaceStatus.Name); err == nil && link.Attrs().EncapType == "ether" {
+				for _, vf := range vfs {
+					if err := Unbind(vf); err != nil {
+						glog.Warningf("configSriovDevice(): fail to unbind driver for device %s", vf)
+					}
+				}
+				if err := setNicSriovMode(iface.PciAddress, iface.SriovMode); err != nil {
```
pliurh: If the hardware doesn't support "switchdev" mode, the `iface.SriovMode` would be empty string, right? If that is the case, I think it would be better to check if `iface.SriovMode` is empty before invoking `setNicSriovMode`.

pkg/plugins/generic/generic_plugin.go
```diff
@@ -103,6 +104,23 @@ func (p *GenericPlugin) OnNodeStateChange(old, new *sriovnetworkv1.SriovNetworkN
...
+			found = false
+			var ifaceOldStatus sriovnetworkv1.InterfaceExt
+			for _, ifaceOldStatus = range old.Status.Interfaces {
+				if ifaceOldStatus.PciAddress == in.PciAddress {
+					found = true
+					break
+				}
+			}
+			if !found || ifaceOldStatus.SriovMode != in.SriovMode {
+				needDrain = true
```
pliurh: Does this change require to drain the node?

pkg/utils/utils.go
```diff
@@ -346,3 +372,47 @@ func Chroot(path string) (func() error, error) {
...
+	pfName, err := dputils.GetPfName(pciAddress)
+	if err != nil {
+		return err
+	}
+	offloadStat := "off"
+	if mode == "switchdev" {
+		offloadStat = "on"
+	}
+	args := []string{"-K", pfName, "hw-tc-offload", offloadStat}
+	_, err = RunCommand("ethtool", args...)
```
pliurh: The `ethtool` may not available in the image. It's better to use `chroot` to run this command from host instead.

pkg/utils/utils.go
```diff
@@ -85,6 +87,12 @@ func DiscoverSriovDevices() ([]sriovnetworkv1.InterfaceExt, error) {
 		if dputils.IsSriovPF(device.Address) {
 			iface.TotalVfs = dputils.GetSriovVFcapacity(device.Address)
 			iface.NumVfs = dputils.GetVFconfigured(device.Address)
+			sriovMode, err := GetNicSriovMode(device.Address)
+			if err != nil {
+				glog.Warningf("DiscoverSriovDevices(): unable to read Sriov mode for device %+v %q", device, err)
+			}
+			iface.SriovMode = sriovMode
+			iface.NumVfs = dputils.GetVFconfigured(device.Address)
```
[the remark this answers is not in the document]
Mmduh-483: By mistake

pkg/utils/utils.go
```diff
@@ -169,6 +177,24 @@ func configSriovDevice(iface *sriovnetworkv1.Interface, ifaceStatus *sriovnetwor
...
+				}
+				if err := setNicSriovMode(iface.PciAddress, iface.SriovMode); err != nil {
+					return err
+				}
+				sriovMode, err := GetNicSriovMode(iface.PciAddress)
+				if err != nil {
+					return err
+				}
+				ifaceStatus.SriovMode = sriovMode
+			}
```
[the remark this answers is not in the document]
Mmduh-483: In case of IB without IPoIB module loaded it will fail, but the check for the ifaceStatus.Name prevents the scenario, I updated it to return an error

pkg/utils/utils.go
Mmduh-483: PF should have VFs created to be able to read or update the sriov mode, some PFs may not have VFs created, I add a check in the DiscoverSriovDevices() but it will fail if PF doesn't have VFs

pkg/utils/utils.go
Mmduh-483: I agree

pkg/utils/utils.go
Mmduh-483: It'd be good to install it in the image by updating the docker file for the config daemon

pkg/utils/utils.go
adrianchiris: for PFs that have no VFs, specifying SriovMode should not be allowed (or blocked if SriovMode is switchdev at the very least)

ConfigSriovDevice() starts with configuring VFs, @Mmduh-483  can you elaborate the scenario where updating the ifaceStatus is required here ?

Mmduh-483:
@adrianchiris Added a check for iface.NumOfVfs > 0 in the generic plugin

pkg/utils/utils.go
adrianchiris: agree with @pliurh  the status needs to be updated in one place (DiscoverSriovDevices). its ok imo to have a short period of time where the reflected node state is not in sync with the actual node state.

pkg/utils/utils.go
```diff
@@ -169,6 +176,28 @@ func configSriovDevice(iface *sriovnetworkv1.Interface, ifaceStatus *sriovnetwor
 		if err != nil {
 			glog.Warningf("configSriovDevice(): unable to parse VFs for device %+v %q", iface.PciAddress, err)
 		}
+		// Set Sriov mode for ETH devices
+		if iface.SriovMode != "" && ifaceStatus.Name != "" {
+			link, err := netlink.LinkByName(ifaceStatus.Name)
+			if err != nil {
+				return err
+			}
+			if link.Attrs().EncapType == "ether" {
```
adrianchiris: do we really need to check the link state?
switchdev/legacy configuration is being done by devlink which is not aware of the link type

pkg/utils/utils.go
Mmduh-483: I removed it from the utils it is now updated from DiscoverSriovDevices

pkg/plugins/generic/generic_plugin.go
Mmduh-483: @adrianchiris moved it to new function

pkg/utils/utils.go
Mmduh-483: @pliurh ethool may not be also in the host

pkg/plugins/generic/generic_plugin.go
```diff
@@ -103,6 +104,27 @@ func (p *GenericPlugin) OnNodeStateChange(old, new *sriovnetworkv1.SriovNetworkN
...
+			}
+			found = false
+			var ifaceOldStatus sriovnetworkv1.InterfaceExt
+			for _, ifaceOldStatus = range old.Status.Interfaces {
+				if ifaceOldStatus.PciAddress == in.PciAddress {
+					found = true
+					break
+				}
+			}
+			if !found || ifaceOldStatus.SriovMode != in.SriovMode {
```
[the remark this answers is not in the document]
Mmduh-483: Agree

pkg/utils/utils.go
Mmduh-483: yes, IB doesn't support changing sriov mode

Mmduh-483:
@adrianchiris @pliurh Any other comments ?

pkg/plugins/generic/generic_plugin.go
```diff
@@ -197,3 +203,24 @@ func isCommandNotFound(err error) bool {
...
+			return false, fmt.Errorf("generic-plugin OnNodeStateChange(): " +
+				"To change SR-IOV mode number of VFs should be > 0")
+		}
+		for _, ifaceOldStatus := range old.Status.Interfaces {
+			if ifaceOldStatus.PciAddress == iface.PciAddress && ifaceOldStatus.SriovMode != iface.SriovMode {
+				needDrain = true
+			}
+		}
+	}
+	return
```
adrianchiris: return needDrain, err ?

pkg/plugins/generic/generic_plugin.go
```diff
@@ -197,3 +203,24 @@ func isCommandNotFound(err error) bool {
 	}
 	return false
 }
+
+func handleSriovModeStateChange(iface *sriovnetworkv1.Interface, old *sriovnetworkv1.SriovNetworkNodeState) (needDrain bool, err error) {
+	err = nil
+	needDrain = false
+	if iface.SriovMode == "" {
```
adrianchiris: if iface.SriovMode **!=** "" ?

pkg/plugins/generic/generic_plugin.go
```diff
@@ -197,3 +203,24 @@ func isCommandNotFound(err error) bool {
...
+		if iface.SriovMode != "legacy" && iface.SriovMode != "switchdev" {
+			return false, fmt.Errorf("generic-plugin OnNodeStateChange(): not supported SR-IOV mode %s for "+
+				"device %s, supported modes [legacy, switchdev]", iface.PciAddress, iface.SriovMode)
+		}
+		if iface.NumVfs <= 0 {
+			return false, fmt.Errorf("generic-plugin OnNodeStateChange(): " +
+				"To change SR-IOV mode number of VFs should be > 0")
```
adrianchiris: "To change SR-IOV mode, number of VFs should be > 0"

pkg/plugins/generic/generic_plugin.go
```diff
@@ -197,3 +203,24 @@ func isCommandNotFound(err error) bool {
 	}
 	return false
 }
+
+func handleSriovModeStateChange(iface *sriovnetworkv1.Interface, old *sriovnetworkv1.SriovNetworkNodeState) (needDrain bool, err error) {
+	err = nil
+	needDrain = false
+	if iface.SriovMode == "" {
+		if iface.SriovMode != "legacy" && iface.SriovMode != "switchdev" {
+			return false, fmt.Errorf("generic-plugin OnNodeStateChange(): not supported SR-IOV mode %s for "+
```
adrianchiris: Unsupported instead of not supported

pkg/plugins/generic/generic_plugin.go
```diff
@@ -197,3 +203,24 @@ func isCommandNotFound(err error) bool {
...
+				"device %s, supported modes [legacy, switchdev]", iface.PciAddress, iface.SriovMode)
```
adrianchiris: can we have consts defined for SriovMode ? you are using "legacy" and "switchdev" strings in several places

pkg/utils/utils.go
```diff
@@ -169,6 +176,23 @@ func configSriovDevice(iface *sriovnetworkv1.Interface, ifaceStatus *sriovnetwor
...
+				for _, vf := range vfs {
+					if err := Unbind(vf); err != nil {
+						glog.Warningf("configSriovDevice(): fail to unbind driver for device %s", vf)
```
adrianchiris: failed

pkg/plugins/generic/generic_plugin.go
Mmduh-483: return values already defined in the function header, it is something supported by go language

pkg/plugins/generic/generic_plugin.go
Mmduh-483: by mistake it should be `if iface.SriovMode != ""`

pkg/plugins/generic/generic_plugin.go
adrianchiris: well, thats what you get when you write in python most of the time now :)

pkg/plugins/generic/generic_plugin.go
Mmduh-483: it only used once as string in the if condition on line 211 and in the error message in error message

adrianchiris: nevertheless i think its best practice to have this defined, also devlink relies on this string, it would lower the chances of errors in the future.

pkg/plugins/generic/generic_plugin.go
adrianchiris: i see, thx for clarifying

Mmduh-483:
@adrianchiris @pliurh All previous comments are fixed now

Mmduh-483:
@moshe010

Mmduh-483:
@pliurh @zshi-redhat Can you please give a review

zshi-redhat:
@Mmduh-483 please split vendor pkg update and actual function change into separate commits.

Mmduh-483:
> @Mmduh-483 please split vendor pkg update and actual function change into separate commits.

done

PR #76, opened 2019-09-10
BUG 1750323: Add a global manager to watch net-att-def in all namespaces

pliurh:


zshi-redhat:
@pliurh This PR adds a new controller for sriovnetwork which seems a relatively big change given that we are close to code freeze date and I'm not sure if QE has enough time to run all the tests. Shall we consider including it in 4.3?

pkg/controller/sriovnetwork/sriovnetwork_controller.go
```diff
@@ -143,35 +126,49 @@ func (r *ReconcileSriovNetwork) Reconcile(request reconcile.Request) (reconcile.
...
 		reqLogger.Info("more than one NetworkAttachmentDefinition CR exists for one SriovNetwork CR", "Namespace", instance.GetNamespace(), "Name", instance.GetName())
-		ns := ""
-		if  instance.Spec.NetworkNamespace != "" {
-			ns = instance.Spec.NetworkNamespace
-		} else {
-			ns = instance.GetNamespace()
-		}
 		for _, nad := range sriovNads {
-			if nad.GetNamespace() != ns {
+			if nad.GetNamespace() != netAttDef.GetNamespace() {
```
zshi-redhat: This may wrongly delete net-attach-def CRs that have same name as defined in SriovNetwork, but in a different namespace. Shall we check CNI type here and only apply when it's `sriov`?
CNO may support creating net-attach-def for other types of CNI. If both try to recover the deleted net-attach-def, will there be confliction between CNO and SRIOV Operator?

pliurh: It will not, there is a check above to only put CR with matched owner ID (which is the ID of the SriovNetwork CR) to `sriovNads`.

pliurh:
> @pliurh As we talked yesterday, this PR adds a new controller for sriovnetwork which seems a relatively big change given that we are close to code freeze date and I'm not sure if QE has enough time to run all the tests. Shall we consider including it in 4.3?

I agree QE need to run some regression tests for this, however I don't think QE need to re-run all the tests. This PR mainly impacts the case related to `SriovNetwork`CRD. I would like to have it merged in 4.2 now, otherwise we may still need to backport it 4.2 anyway.

pkg/controller/sriovnetwork/sriovnetwork_controller.go
zshi-redhat: Yes, I didn't notice there is a loop to filter net-attach-def created by sriovnetwork.

PR #89, opened 2019-10-09
Upgrade operator-sdk to v0.11.0

pliurh:


zshi-redhat:
Is there doc explaining which operator-sdk version shall be used for exposing metrics?

pliurh:
> Is there doc explaining which operator-sdk version shall be used for exposing metrics?

Actually, we don't need this PR to expose the metrics. However, in operator-sdk 0.9.0, there are some changes in the metric server provided by the operator-sdk. So I think it would be better if we keep the operator-sdk up-to-date.

aneeshkp:
Right , there is not required to upgrade to latest to support metrics but the latest have more support  .BTW  The latest  is 0.11.0 , why you upgraded to 0.9.0 only ?

pliurh:
@aneeshkp When I draft this PR, in this [doc](https://github.com/operator-framework/operator-sdk/blob/master/doc/migration/version-upgrade-guide.md), there was no instruction on how to upgrade to 0.11.0. But now, there is. So I will update this PR for 0.11.0.

PR #91, opened 2019-10-11
add manifests/art.yaml

sosiouxme:
This gives us different CSVs for different releases (e.g. 4.2 and 4.3)
when we're building OCP downstream.

This should be cherry-picked back to 4.2 but the problem we have (duplicate CSVs) will go away for now even if just 4.3 gets changed.

### DIFF
--- a/manifests/art.yaml
+++ b/manifests/art.yaml
@@ -1,0 +1,12 @@
+updates:
+  - file: "{MAJOR}.{MINOR}/sriov-network-operator.v0.0.1.clusterserviceversion.yaml" # relative to this file
+    update_list:
+    # replace metadata.name value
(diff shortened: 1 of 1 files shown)


pliurh:
@sosiouxme Is that mandatory to backport this PR to 4.2? Shall I wait for 4.2.z to do the backporting?

sosiouxme:
@pliurh thanks! I am not certain what the consequence of leaving the version as `v0.0.1` for 4.2 is. I doubt it's urgent / mandatory, but it would be clearer and consistent with what the other operators do (especially regarding upgrades). I'm pretty sure at this point though that it should wait for 4.2.z :)

PR #92, opened 2019-10-15
Add udev rule to prevent NetworkManager from managing VFs

pliurh:


pkg/apis/sriovnetwork/v1/helper.go
```diff
@@ -13,6 +13,8 @@ var SriovPfVfMap = map[string](string){
 	"1017": "1018",
 }
 
+var VfIds = []string{"0x154c", "0x10ed", "0x1016", "0x1018"}
```
zshi-redhat: nit: what about using `SriovPfVfMap` to get the supported VF IDs? then we don't need update two places when adding new NIC support.

pliurh: fixed.

PR #96, opened 2019-10-17
Add admission controller webhook for the Operator API

pliurh:


hack/env.sh
```diff
@@ -9,3 +9,4 @@ export SRIOV_NETWORK_CONFIG_DAEMON_IMAGE=${SRIOV_NETWORK_CONFIG_DAEMON_IMAGE:-qu
 export RELEASE_VERSION=0.0.1-snapshot
 export OPERATOR_NAME=sriov-network-operator
 export RESOURCE_PREFIX=openshift.io
+export ENABLE_ADMISSION_CONTROLLER='true'
```
fedepaol: Doesn't this enable the resources mutating webhook (and only when running locally) ?
What is its relation with these new webhooks?

zshi-redhat: With https://github.com/openshift/sriov-network-operator/pull/94, Operaoter controls SRIOV admission controller (resource injector) via `SriovOperatorConfig` CRD (spec.EnableInjector). Env var (ENABLE_ADMISSION_CONTROLLER) in this PR is going to enable/disable for the Operator's own admission controller which is used to mutate/validate Operator's CRD.

zshi-redhat:
> I am probably missing something, but how this new webhook is launched?
> Should it be handled by the operator the same way the network injector (optionally) is? Will this part of a follow-up pr?

I think there will be another PR to handle rdnering and reconciling of this webhook. (not yet posted)

hack/env.sh
fedepaol: Oh, I see.

cmd/webhook/start.go
```diff
@@ -0,0 +1,210 @@
...
+
+	keyPair, err := webhook.NewTlsKeypairReloader(certFile, keyFile)
+	if err != nil {
+		glog.Fatalf("error load certificate: %s", err.Error())
+	}
+
+
+	http.HandleFunc("/mutating-custom-resource", serveMutateCustomResource)
+	http.HandleFunc("/validating-custom-resource", serveValidateCustomResource)
+	http.HandleFunc("/readyz", func(w http.ResponseWriter, req *http.Request) { w.Write([]byte("ok")) })
```
zshi-redhat: what's this handler used for?

pliurh: This is for `readinessProbe` of the pod, which can be used to detect if a pod is ready for service.

PR #100, opened 2019-10-21
Sriov operator config new

zshi-redhat:
Sriov Operator Config API exposes cluster-wide config
option such as EnableInjector, which is used to enable/
disable SR-IOV admission controller.


pkg/controller/sriovoperatorconfig/sriovoperatorconfig_controller.go
```diff
@@ -0,0 +1,282 @@
...
+// add adds a new Controller to mgr with r as the reconcile.Reconciler
+func add(mgr manager.Manager, r reconcile.Reconciler) error {
+	// Create a new controller
+	c, err := controller.New("sriovoperatorconfig-controller", mgr, controller.Options{Reconciler: r})
+	if err != nil {
+		return err
+	}
+
+	// Watch for changes to primary resource SriovOperatorConfig
+	err = c.Watch(&source.Kind{Type: &sriovnetworkv1.SriovOperatorConfig{}},
```
pliurh: I think it would be better to also watch the daemonset resource here.

pkg/controller/sriovoperatorconfig/sriovoperatorconfig_controller.go
```diff
@@ -0,0 +1,282 @@
...
+			err = r.client.Create(context.TODO(), in)
+			if err != nil {
+				return fmt.Errorf("Couldn't create config map: %v", err)
+			}
+			logger.Info("Create config map for", in.Namespace, in.Name)
+		} else {
+			return fmt.Errorf("Fail to get config map: %v", err)
+		}
+	}
+
```
pliurh: I think we shall make sure the annotation of the CM is intact. Update it to the correct value when needed.

zshi-redhat: do you mean update the CM here? (I don't know if there is a way to just update annotation)

pliurh: You can just reuse the object `cm` from `Get` call, and modify the annotation when need.

pkg/controller/sriovoperatorconfig/sriovoperatorconfig_controller.go
zshi-redhat: fixed

pkg/controller/sriovoperatorconfig/sriovoperatorconfig_controller.go
zshi-redhat: fixed

PR #106, opened 2019-10-25
Allow custom nodeSelector for config daemon

pliurh:


zshi-redhat:
If `network config daemon` pod is removed from a node, how will it affect the running pod on that node?

zshi-redhat:
Shall `SriovNetworkNodeState` be created only on nodes that run config daemon?

pliurh:
> Shall `SriovNetworkNodeState` be created only on nodes that run config daemon?

good point.

pliurh:
> Shall `SriovNetworkNodeState` be created only on nodes that run config daemon?



> If `network config daemon` pod is removed from a node, how will it affect the running pod on that node?

I don't think there will be any.

PR #111, opened 2019-10-31
Operator Webhook: Validate max allowed VFs for Mellanox nics

pliurh:


pkg/webhook/validate.go
```diff
@@ -78,7 +80,7 @@ func validatePolicyForNodeState(policy *sriovnetworkv1.SriovNetworkNodePolicy, s
 	for _, iface := range state.Status.Interfaces {
 		if policy.Spec.Selected(&iface) {
 			interfaceSelected = true
-			if policy.Spec.NumVfs > iface.TotalVfs && iface.Vendor == IntelID {
+			if (policy.Spec.NumVfs > iface.TotalVfs && iface.Vendor == IntelID) || (policy.Spec.NumVfs > MlxMaxVFs && iface.Vendor == MellanoxID) {
 				return false, fmt.Errorf("numVfs(%d) in CR %s exceed the maximum allowed value", policy.Spec.NumVfs, policy.GetName())
```
zshi-redhat: Can we print the maxVF number here once conditional check fails?
This will help user to know what number should be configured for a certain device.

pliurh: fixed

PR #112, opened 2019-11-01
E2E: wait for operator webhook up  before testing NodePolicy creation

pliurh:


test/e2e/e2e_test.go
```diff
@@ -99,7 +99,9 @@ func TestOperatorController(t *testing.T) {
 			})
 		}
 	})
-
+	
+	// wait for operator webhook endpoint up and running
+	time.Sleep(30 * time.Second)
```
zshi-redhat: Is it possible to use `oc wait --for condition=`  waiting daemonset be ready or available?

PR #114, opened 2019-11-06
Update maxTxRate and minTxRate change to Operator bundle

pliurh:


zshi-redhat:
do we need a new BZ for this PR?

PR #116, opened 2019-11-07
Allow to disable node reboot

mmirecki:
If the SriovOperatorConfig contains disableReboot=true
the sriov-network-config-daemon will not reboot the nodes
but instead just log this and exit.
(the pr was orphaned by github, replaced by PR 229)


cmd/sriov-network-config-daemon/start.go
```diff
@@ -87,6 +88,10 @@ func runStartCmd(cmd *cobra.Command, args []string) {
 
 	snclient := snclientset.NewForConfigOrDie(config)
 	kubeclient := kubernetes.NewForConfigOrDie(config)
+	dynamicclient, err := dynamic.NewForConfig(config)
```
fedepaol: Can you use NewForConfigOrDie (as per the others)?
Also, can the name be `dynamicClient`?

pkg/daemon/daemon.go
```diff
@@ -70,17 +76,19 @@ func New(
 	nodeName string,
 	client snclientset.Interface,
 	kubeClient *kubernetes.Clientset,
+	dynamicclient *dynamic.Interface,
```
fedepaol: `dynamicClient`

pkg/daemon/daemon.go
```diff
@@ -515,3 +533,24 @@ func tryCreateUdevRule() error {
...
 }
+
+func (dn *Daemon) rebootDisabled() bool {
+
+	deploymentRes := schema.GroupVersionResource{
+		Group:    "sriovnetwork.openshift.io",
+		Version:  "v1",
+		Resource: "sriovoperatorconfigs",
+	}
+	sriovoperatorconfig, err := dn.dynamicclient.Resource(deploymentRes).Namespace("default").Get("default", metav1.GetOptions{})
```
fedepaol: camel case: `sriovOperatorConfig`

pkg/daemon/daemon.go
```diff
@@ -515,3 +533,24 @@ func tryCreateUdevRule() error {
...
+	if err != nil {
+		glog.V(2).Infof("rebootDisabled(): Failed to check disableReboot: SriovOperatorConfig not found.")
+		return false
+	}
+	spec := sriovoperatorconfig.Object["spec"].(map[string]interface{})
+	isreboot, ok := spec["disableReboot"]
```
fedepaol: Can it be `disableReboot`?

pkg/daemon/daemon.go
```diff
@@ -515,3 +533,24 @@ func tryCreateUdevRule() error {
...
+	if !ok {
+		glog.V(2).Infof("rebootDisabled(): Failed to check disableReboot: SriovOperatorConfig does not contain disableReboot.")
+		return false
+	}
+	return isreboot.(bool)
```
fedepaol: Can you also check that the variable is really a bool (otherwise it will panic)?
res, ok := isreboot.(bool)

pkg/daemon/daemon.go
fedepaol: Also, fetching the value each time is not that efficient but since this does not happen often I think it's more than acceptable.

mmirecki: rebootDisabled() is only invoked once when the daemon starts

fedepaol: even better. Sorry I was thinking it was invoked every time it needed to be checked.

cmd/sriov-network-config-daemon/start.go
mmirecki: Done

pkg/daemon/daemon.go
mmirecki: Done

pkg/daemon/daemon.go
mmirecki: Done

pkg/daemon/daemon.go
mmirecki: Done

pkg/daemon/daemon.go
mmirecki: Done

pliurh:
/hold
Since we have passed the 4.3 feature freeze, this PR can only be merged when 4.4 branch is open.

pkg/daemon/daemon.go
```diff
@@ -185,6 +195,10 @@ func (dn *Daemon) nodeStateAddHandler(obj interface{}) {
 	}
 
 	if reqReboot {
+		if dn.noReboot {
```
pliurh: I think it would be better to invoke `rebootDisabled()` before executing the reboot. Since the SriovOperatorConfig CR may be changed after the config daemon is up.

pkg/daemon/daemon.go
```diff
@@ -291,6 +305,10 @@ func (dn *Daemon) nodeStateChangeHandler(old, new interface{}) {
...
```
pliurh: The same as above comment

PR #118, opened 2019-11-11
BUG 1770642: Set PF MTU when deviceType is vfio

pliurh:


pkg/utils/utils.go
```diff
@@ -230,16 +230,15 @@ func setSriovNumVfs(pciAddr string, numVfs int) error {
...
-	ifaceName, err := dputils.GetNetNames(pciAddr)
-	if err != nil {
-		glog.Warningf("setNetdevMTU(): fail to get interface name for %s: %s", pciAddr, err)
-		return err
-	}
-	mtuFile := "net/" + ifaceName[0] + "/mtu"
-	mtuFilePath := filepath.Join(sysBusPciDevices, pciAddr, mtuFile)
 	b := backoff.NewConstantBackOff(1 * time.Second)
-	err = backoff.Retry(func() error {
+	err := backoff.Retry(func() error {
```
zshi-redhat: could you explain why above lines are moved into `backoff.Retry()`?

pliurh: Sometimes, after VFs are created, the sysfs is not updated immediately. In that case, the `dputils.GetNetNames()` call will failed with `file/directory not exist`. Therefore I move it into the retry cycle.

PR #119, opened 2019-11-11
BUG 1770709: Use "preserveUnknownFields: false" for the CRD

pliurh:


Makefile
```diff
@@ -102,3 +102,13 @@ else
 endif
 
 verify: verify-gofmt
+
+sync-manifests:
```
zshi-redhat: This may not work for 4.3 if there is API change in 4.4 development cycle. where only 4.4 needs to be updated.

pliurh: Fiixed

PR #126, opened 2019-11-26
Bug 1774419: Webhook Add validation for resourceName

pliurh:


zshi-redhat:
Just to double confirm, do we watch net-attach-def creation and validate resource name in it in operator webhook?

pliurh:
> Just to double confirm, do we watch net-attach-def creation and validate resource name in it in operator webhook? I don;t think we do, right?

No, we don't. Currently, the operator webhook only validates the SriovNetworkNodePolicy CR.

PR #138, opened 2019-12-16
Support specifying VF index in pfNames

### LINKED ISSUE #88: RFE: allow to split VFs of the same PF into multiple resource pools
Use case: a node has a single PF with multiple VFs. An admin would like to allow scheduling different workloads to the same node. Specifically, the admin would like the cluster users to schedule both workloads that rely on VFIO (e.g. kubevirt VMIs) as well as those that use the regular network driver (regular pods using netlink interfaces).

Problem: right now, the operator spec format doesn't allow to specify that e.g. first 10 VFs should be configured for VFIO while the rest left for `netdevice` mode of the device plugin.

Related info: the device plugin now allows to list indices of VFs when using PF name selector: https://github.com/intel/sriov-network-device-plugin/issues/157 But since there are issues with requiring listing particular PF names in the spec + not being able to select a particular node, the new feature may not rely on explicit name/indices.

Related discussion: there was a discussion among operator maintainers and kubevirt engineers (via email) with ideas about how this feature could be implemented that I will capture below for reference.

====

If we want to support it, I prefer to minimize modification to the API. Here is a thought. Please see if it can fulfil your use case.

Currently, there is priority field in the SriovNetworkNodePolicy CR. When there are multiple policies select the same set of SRIOV NICs, and specify different numVFs, the higher priority policy will overruled the lower ones. Maybe we can add a "sharedPF" field to overlook this priority behavior, when it's true, operator sums up the total numVFs in all polices, then config the PF accordingly.

In following example, there're 2 policies defined for 2 resource names. We can let operator to config 6+6=12 VFs for NIC "eno3", then load vfio driver for 6 VFs, and load default driver for the rest. And in the config of device plugin, we can have two resourceNames which has different "driver selector".

```
---
apiVersion: sriovnetwork.openshift.io/v1
kind: SriovNetworkNodePolicy
metadata:
  name: policy-1
  namespace: sriov-network-operator
spec:
  resourceName: intelnics
  nodeSelector:
    feature.node.kubernetes.io/network-sriov.capable: "true"
  priority: 99
  sharedPF: true
  numVfs: 6
  nicSelector:
    vendor: "8086"
    pfName: eno3
    rootDevices: ['0000:86:00.1']
  deviceType: netdevice
---
apiVersion: sriovnetwork.openshift.io/v1
kind: SriovNetworkNodePolicy
metadata:
  name: policy-2
  namespace: sriov-network-operator
spec:
  resourceName: intelnics-dpdk
  nodeSelector:
    feature.node.kubernetes.io/network-sriov.capable: "true"
  priority: 99
  sharedPF: true
  numVfs: 6
  nicSelector:
    vendor: "8086"
    pfName: eno3
    rootDevices: ['0000:86:00.1']
  deviceType: vfio-pci
```

===

Just to share another thought on splitting VFs from same PF as different resource names.
There is a open PR in device plugin to enhance PF selector to specify VF index, for example:
PF filter will be like ["eth0#1-3"] indicating that 3 VFs will need to be created from eth0 PF and VFs with index 1 to 3 will be grouped as a single resource name. I think SR-IOV Operator might be able to leverage this feature, inspect the PF filter to get VF index numbers before merging policies, then decide on how many VFs should be provisioned on each PF.
This has advantage of not changing any operator APIs and allows use of VFs from same PF as different resources even they have same driver, but it requires PF selector(optional) be configured in order to split VFs from same PF.

https://github.com/intel/sriov-network-device-plugin/pull/165

===

I think the original asking is to group VF devices with different types (netdevice or vfio-pci) to separate resource pools.
This is already supported in DP by using driver selector. I don't think we need to add another selector in DP to support this particular use case.

The missing parts are:
1) we need to find a way in SR-IOV Operator to generate such configuration for DP.
2) operator cannot provision single PF to VFs of different device types. Because according to SR-IOV node network policy design, only one policy with highest priority matching PF device will be applied. If device type is specified as vfio-pci, then all provisioned VFs will be bind to vfio-pci driver.

The suggestion of using VF index in policy is to allow Operator to configure VFs from same PF differently. Operator can have its own format of defining VF index, for example, it may be appended to each element of rootDevices list which is used to identify PF device and provision VFs on it. This could solve the second issue.

For the 1) issue, if PF name selector is not configured in SR-IOV node network policy (final merged policy doesn't contain PF name selector), then generated DP configuration won't contain PF name.

pliurh:
This PR allows the SR-IOV Network Operator to support the SR-IOV device plugin feature discussed in https://github.com/intel/sriov-network-device-plugin/issues/157. This allows user to specify the VF range in SriovNetworkNodePolicy CR with fomat `"<PFName>#<FirstVF>-<LastVF>"` in `pfName` field.

This PR also adds a `vfGroups` field to the SriovNetworkNodeState CRD, so that the configure daemon can use it as input to configure the VFs which are listed in the vfGroups. For the rest VFs, the operator will trade them as 'unmanaged' VFs. The 'unmanaged' VFs may be used by the host for other purposes.

close #88


zshi-redhat:
This may unlock the use case that VFs from the same PF can be used with different drivers.

pliurh:
> This may unlock the use case that VFs from the same PF can be used with different drivers.

Yes.

dankenigsberg:
Would you add introductory text explaining the motivation of this pr?

Also, I failed to find intree documentation of the API change

PR #144, opened 2020-01-30
Bump master to version 4.5

pliurh:


manifests/sriov-network-operator.package.yaml
```diff
@@ -1,4 +1,4 @@
 packageName: sriov-network-operator
 channels:
-- name: "4.4"
-  currentCSV: sriov-network-operator.v4.4.0
+- name: "4.5"
```
zshi-redhat: Should this include the latest channel or all available channels?

manifests/art.yaml
```diff
@@ -10,5 +10,5 @@ updates:
       replace: 'olm.skipRange: ">=4.3.0-0 <{FULL_VER}"'
   - file: "sriov-network-operator.package.yaml"
     update_list:
-    - search: "currentCSV: sriov-network-operator.v4.4.0"
+    - search: "currentCSV: sriov-network-operator.v4.5.0"
```
zshi-redhat: Perhaps change this to `currentCSV: ptp-operator.v{MAJOR}.{MINOR}.0` so that we don't need to update the version for every release.

pliurh: Fixed.

manifests/sriov-network-operator.package.yaml
pliurh: I think ART will create images for each release. Here we shall only put the latest one.

PR #146, opened 2020-02-03
BUG 1798043: Optimize VF configure logic

pliurh:
Check whether desired state is the same as the current one before
executing.

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -131,29 +131,37 @@ func SyncNodeState(newState *sriovnetworkv1.SriovNetworkNodeState) error {
 }
 
 func needUpdate(iface *sriovnetworkv1.Interface, ifaceStatus *sriovnetworkv1.InterfaceExt) bool {
-	if iface.Mtu != ifaceStatus.Mtu {
+	if iface.Mtu > 0 && iface.Mtu != ifaceStatus.Mtu {
+		glog.V(2).Infof("needUpdate(): MTU needs update, desired=%d, current=%d", iface.Mtu, ifaceStatus.Mtu)
 		return true
 	}
 	if iface.NumVfs != ifaceStatus.NumVfs {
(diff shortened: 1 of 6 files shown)


pkg/utils/utils.go
```diff
@@ -131,54 +131,66 @@ func SyncNodeState(newState *sriovnetworkv1.SriovNetworkNodeState) error {
...
 		glog.Errorf("configSriovDevice(): fail to set NumVfs for device %s: %v", iface.PciAddress, err)
 		return err
 	}
-
-	err := setSriovNumVfs(iface.PciAddress, iface.NumVfs)
-	if err != nil {
-		glog.Errorf("configSriovDevice(): fail to set NumVfs for device %s", iface.PciAddress)
-		return err
+	// set numVFs
+	if iface.NumVfs != ifaceStatus.NumVfs {
```
zshi-redhat: This is checked in `needUpdate`,  is there other reason to check it here again?

pkg/utils/utils.go
```diff
@@ -131,54 +131,66 @@ func SyncNodeState(newState *sriovnetworkv1.SriovNetworkNodeState) error {
...
+			glog.Errorf("configSriovDevice(): fail to set NumVfs for device %s", iface.PciAddress)
+			return err
+		}
 	}
 	// set PF mtu
-	err = setNetdevMTU(iface.PciAddress, iface.Mtu)
-	if err != nil {
-		glog.Warningf("configSriovDevice(): fail to set mtu for PF %s: %v", iface.PciAddress, err)
-		return err
+	if iface.Mtu > 0 && iface.Mtu != ifaceStatus.Mtu {
```
zshi-redhat: ditto

pkg/utils/utils.go
pliurh: needUpdate will return 'True' if any of the parameter has been updated. However, we need to check if one parameter shall be updated separately before executing.

PR #147, opened 2020-02-05
[release-4.3] BUG 1798344: Optimize VF configure logic

pliurh:
Check whether desired state is the same as the current one before
executing.

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -131,10 +131,12 @@ func SyncNodeState(newState *sriovnetworkv1.SriovNetworkNodeState) error {
 }
 
 func needUpdate(iface *sriovnetworkv1.Interface, ifaceStatus *sriovnetworkv1.InterfaceExt) bool {
-	if iface.Mtu != ifaceStatus.Mtu {
+	if iface.Mtu > 0 && iface.Mtu != ifaceStatus.Mtu {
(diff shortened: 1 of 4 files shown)


pkg/utils/utils.go
```diff
@@ -131,10 +131,12 @@ func SyncNodeState(newState *sriovnetworkv1.SriovNetworkNodeState) error {
 }
 
 func needUpdate(iface *sriovnetworkv1.Interface, ifaceStatus *sriovnetworkv1.InterfaceExt) bool {
-	if iface.Mtu != ifaceStatus.Mtu {
+	if iface.Mtu > 0 && iface.Mtu != ifaceStatus.Mtu {
```
zshi-redhat: The range of `iface.Mtu` is defined in crd policy definition, do we still need to check '>0' ? Similar to `iface.NumVfs`, we didn't check if it's larger than 0.

pliurh: The MTU can be omitted in policy, in such case, golang will assign a default value '0' to the int variable.  So if user doesn't specify the MTU in policy explicitly, the configure daemon shall not set the MTU. 

For NumVfs, we shall reset the PF to 0 VF, when user set the NumVfs to 0 in policy either explicitly or implicitly.

PR #152, opened 2020-02-14
Bug 1802936: Set default vfRange when pfName is not specified in policy

pliurh:


zshi-redhat:
The patch looks good to me, it might be good to add some code annotation for easy reading later.

pliurh:
> The patch looks good to me, it might be good to add some code annotation for easy reading later.

Fixed.

PR #155, opened 2020-02-21
Bug 1798880: Prevent user from configuring overlapped VfRange

pliurh:


pkg/webhook/validate.go
```diff
@@ -135,5 +135,27 @@ func validatePolicyForNodeState(policy *sriovnetworkv1.SriovNetworkNodePolicy, s
 			}
 		}
 	}
+
+	for _, iface := range state.Spec.Interfaces {
```
zshi-redhat: Shall we put the same logic in operator in case webhook is disabled?

pliurh: Currently, in the operator, all the policies will be applied by the order of their priorities, then by the alphabet order of their name when the priorities are identical. This means if there is any overlapping among the policies, the policy applied later can override the one applied earlier. 

I think this behavior is rational, as long as the user knows what they are doing when disabling the webhook. Otherwise, users shall not disable the webhook, as that is meant to be used to prevent such unintentional configuration conflict.

zshi-redhat: > Currently, in the operator, all the policies will be applied by the order of their priorities, then by the alphabet order of their name when the priorities are identical. This means if there is any overlapping among the policies, the policy applied later can override the one applied earlier.

I don't see how the problem can be mitagated by applying order. do you mean the overlapped resource will only be counted in the last applied policy/resource?

> 
> I think this behavior is rational, as long as the user knows what they are doing when disabling the webhook. Otherwise, users shall not disable the webhook, as that is meant to be used to prevent such unintentional configuration conflict.

I think we provide the flag to disable webhook and should not assume all the user configuration will be applied with it enabled. If fixing this in operator code is not appropriate, it might be worth a note in the document where operator webhook is explained.

pliurh: > > Currently, in the operator, all the policies will be applied by the order of their priorities, then by the alphabet order of their name when the priorities are identical. This means if there is any overlapping among the policies, the policy applied later can override the one applied earlier.
> 
> I don't see how the problem can be mitagated by applying order. do you mean the overlapped resource will only be counted in the last applied policy/resource?

I don't mean the problem can be mitigated. I just described how the policies will be applied if users turn off the webhook and configure overlapping policies. So the message is that users shall NOT turn off the webhook unless they know exactly what will happen.

> 
> > I think this behavior is rational, as long as the user knows what they are doing when disabling the webhook. Otherwise, users shall not disable the webhook, as that is meant to be used to prevent such unintentional configuration conflict.
> 
> I think we provide the flag to disable webhook and should not assume all the user configuration will be applied with it enabled. If fixing this in operator code is not appropriate, it might be worth a note in the document where operator webhook is explained.

We have stated in the document that users shall not create policies with overlapping VF index.  Again, I don't think we shall tell users not to disable the webhook unless doing debug.

zshi-redhat:
We believe that with current fix, it would cover most cases where operator webhook is enabled. It'd be rare that user disables operator webhook and configures overlapped vf ranges.
If the issue does come up later with operator webhook disabled, we will consider fixing it in operator code as well as device plugin code.

/lgtm

PR #156, opened 2020-02-24
[release-4.3] Bug 1806376: Backport pf partitioning to 4.3

pliurh:


zshi-redhat:
Is this commit `7a484bb Fix api version for daemonsets` needed for 4.3? I think we could still use old API version `extentions`.
Do we need to backport SR-IOV Network fix: https://bugzilla.redhat.com/show_bug.cgi?id=1770668 to 4.3?

zshi-redhat:
what about `e25312b Update sriov-network-operator-sriovnetworknodestate.crd.yaml`? It was fixed in the same PR of SR-IOV Partitioning for 4.4.

pliurh:
> Is this commit `7a484bb Fix api version for daemonsets` needed for 4.3? I think we could still use old API version `extentions`.
> Do we need to backport SR-IOV Network fix: https://bugzilla.redhat.com/show_bug.cgi?id=1770668 to 4.3?

This API has already been available since k8s 1.6. So this fix shall also apply to in 4.3.

pliurh:
> what about `e25312b Update sriov-network-operator-sriovnetworknodestate.crd.yaml`? It was fixed in the same PR of SR-IOV Partitioning for 4.4.

We cannot cherry-pick that commit, but yes, we need to do the same thing for 4.3. I've added a new commit for this.

PR #158, opened 2020-02-27
WIP: Allow user to modify the log level of config daemon on live

pliurh:
The log level of config daemon used to be set through the CLI '-v' flag with a default value '3'. This PR lets users be able to modify the the log level through ConfigMap 'sriov-network-config-daemon' on live.


fedepaol:
@pliurh  As this is going to be read from all the daemons, why not use the sriov operator config CRD ?

pliurh:
> @pliurh As this is going to be read from all the daemons, why not use the sriov operator config CRD ?

Actually, I'm still evaluating how such functionality shall be exposed to users. From my point of view, as it's a feature for debugging, I prefer to make it a somewhat 'hidden' option and not expose it through the API. The concern also applies to the functionality like, disable reboot/drain node. I would like to hear your thoughts. @zshi-redhat

fedepaol:
> > @pliurh As this is going to be read from all the daemons, why not use the sriov operator config CRD ?
> 
> Actually, I'm still evaluating how such functionality shall be exposed to users. From my point of view, as it's a feature for debugging, I prefer to make it a somewhat 'hidden' option and not expose it through the API. The concern also applies to the functionality like, disable reboot/drain node. I would like to hear your thoughts. @zshi-redhat @fedepaol

Ah, makes sense and I agree, if we want this to be an hidden knob, the configmap is a better option. 
I have another question then :-)
Why mapping it to a file and watching it with Viper, as opposed to watching directly the ConfigMap object? 
And also, if we want it to be hidden, wouldn't it make sense to have the default value applied with no ConfigMap at all (i.e with missing file)? 
In that case, you'd have to know what config to create which would make it even more hidden.

zshi-redhat:
> > @pliurh As this is going to be read from all the daemons, why not use the sriov operator config CRD ?
> 
> Actually, I'm still evaluating how such functionality shall be exposed to users. From my point of view, as it's a feature for debugging, I prefer to make it a somewhat 'hidden' option and not expose it through the API. The concern also applies to the functionality like, disable reboot/drain node. I would like to hear your thoughts. @zshi-redhat @fedepaol

I think the inital thought of having this dynamic debug level control is to reduce the detailed message output(such as function call entry) from config daemon for easy debugging and at the same time keep it as an option to enable advanced debugging output if we need to collect as much details as possible for further debugging. So we may still need to ask user to enable this function where needed. 

For diabling reboot/drain nodes, do we need to expose dynamic control over these options or is it sufficient to configure them in cmd option?

pliurh:
Deprecated by #164

PR #164, opened 2020-03-10
Add flag 'logLevel' in SriovOperatorConfig

pliurh:
Flag 'logLevel' is used to set the verbose level of the config daemon.
Currently, the default value is 2 to show all the available log messages.


pliurh:
@zshi-redhat @fedepaol /PTAL

After discussing with @zshi-redhat we decided to expose this flag in from the SriovOperatorConfig API. This patch is only to enable the operator with a dynamic logLevel. I'll groom all the log messages in the config daemon to make the log tidier in separate patches.

pkg/daemon/daemon.go
```diff
@@ -139,8 +170,8 @@ func (dn *Daemon) nodeStateAddHandler(obj interface{}) {
 	err := dn.loadVendorPlugins(nodeState)
 	if err != nil {
 		glog.Errorf("nodeStateAddHandler(): failed to load vendor plugin: %v", err)
-		dn.exitCh <- err
-		return
+		// dn.exitCh <- err
```
fedepaol: Is commenting this lines intentional?

pkg/daemon/daemon.go
```diff
@@ -127,6 +143,21 @@ func (dn *Daemon) Run(stopCh <-chan struct{}, exitCh <-chan error) error {
...
+func (dn *Daemon) operatorConfigAddHandler(obj interface{}) {
+	dn.operatorConfigChangeHandler(&sriovnetworkv1.SriovOperatorConfig{}, obj)
+}
+
+func (dn *Daemon) operatorConfigChangeHandler(old, new interface{}) {
+	oldCfg := old.(*sriovnetworkv1.SriovOperatorConfig)
+	newCfg := new.(*sriovnetworkv1.SriovOperatorConfig)
+
+	if newCfg.Spec.LogLevel != oldCfg.Spec.LogLevel {
+		glog.Infof("Set log verbose level to: %d", newCfg.Spec.LogLevel)
```
fedepaol: Wouldn't it be better to set a global variable (or a field of `Daemon`) to hold the verb level?
Also, I think that this is going to be written / read by different goroutines, as the two informers are spawned separately.

pliurh: There is a global variable in glog to hold the verb level, I think we can just use it. Also, the set flag call in glog it is thread-safe.

pkg/daemon/daemon.go
pliurh: fixed

zshi-redhat:
Looks like we don't need to care about the value of glog level as long as it's an integer, it accepts and handles negative value properly. 

I don't have prefrence on whether using informer or a configmap to pass debug log level to daemons as currently only sriov config daemon is controlled via this flag and there is no need for other components to read it. Going forward, if it's worth adding log level control over other components, it maybe considerable to put it in a configmap that can be read by all daemons instead of having an informer in each daemon. Or just putting it as a variable in bindata that can be reconciled (but this has defect that it needs reboot of daemons to take effect)

/lgtm

PR #166, opened 2020-03-19
Bug 1814919: Prevent user from setting numVfs to '0'

pliurh:
Also, sync the 4.5 CRD in the manifest directory, as it was not done in the operator-sdk upgrade PR.

### DIFF
--- a/manifests/4.5/sriov-network-operator-sriovnetworknodestate.crd.yaml
+++ b/manifests/4.5/sriov-network-operator-sriovnetworknodestate.crd.yaml
@@ -21,12 +21,12 @@ spec:
         apiVersion:
           description: 'APIVersion defines the versioned schema of this representation
             of an object. Servers should convert recognized schemas to the latest
-            internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#resources'
+            internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
           type: string
         kind:
           description: 'Kind is a string value representing the REST resource this
             object represents. Servers may infer this from the endpoint the client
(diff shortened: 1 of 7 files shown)


manifests/4.5/sriov-network-operator-sriovnetworknodestate.crd.yaml
```diff
@@ -108,25 +106,28 @@ spec:
                     type: string
                   linkSpeed:
                     type: string
+                  mac:
```
zshi-redhat: why there is a `mac` field? do we add this in nodestate somewhere?

manifests/4.5/sriov-network-operator-sriovnetworknodestate.crd.yaml
```diff
@@ -92,14 +86,18 @@ spec:
                         mac:
                           type: string
                         mtu:
-                          format: int64
```
zshi-redhat: why this is removed?

pliurh: This CRD file is generated by the latest version operator-sdk. I guess `format` is not necessary.

manifests/4.5/sriov-network-operator-sriovnetworknodestate.crd.yaml
pliurh: It is here https://github.com/openshift/sriov-network-operator/blob/160307f84f7f91d9cd193d063f11350d909617a8/pkg/apis/sriovnetwork/v1/sriovnetworknodestate_types.go#L35
But, currently, we didn't use this parameter at all.

zshi-redhat: shall we remove it if not used?

pliurh: I think we can keep it for now. I plan to implement this function later.

PR #169, opened 2020-03-19
Adding bond-cni install

mmirecki:
This will install the bond-cni plugin executable on the nodes.

This is on HOLD!
The bond-cni repo and CI is not ready yet.
This PR could still be replaced by installing the bond-cni as part of the cluster-network-operator.


zshi-redhat:
> This will install the bond-cni plugin executable on the nodes.
> 
> This is on HOLD!
> The bond-cni repo and CI is not ready yet.

Yes, we need to wait until the CI start building bond-cni image

> This PR could still be replaced by installing the bond-cni as part of the cluster-network-operator.

It's not yet clear whether bond-cni will be used for devices other than SR-IOV.
For a virtual interface that wants to use bonding, there is a problem that the virtual interface may not follow the link state of physical interface it was created on, which cause bond-cni useless.

mmirecki:
Replaced by:  #199

PR #170, opened 2020-03-20
Move e2e and operator tests from sirov-tests repo

pliurh:


fedepaol:
@pliurh you have to run go fmt against a couple of files.
Also, any particular reason why you did not move conformance together with this? 

Other than that, looks good.

pliurh:
@


> @pliurh you have to run go fmt against a couple of files.
forget to do that
> Also, any particular reason why you did not move conformance together with this?
I cannot run all the conformance tests successfully in my env. Would you mind to prepare another PR for that after this PR merged?
> 
> Other than that, looks good.

PR #171, opened 2020-03-23
Revert "Bug 1814919: Prevent user from setting numVfs to '0'"

pliurh:
Reverts openshift/sriov-network-operator#166


zshi-redhat:
revert this commit in favor of having a complete fix in both operator webhook and operator.

/lgtm

PR #175, opened 2020-03-31
BUG 1815039: Retrieve the latest status of SriovNetworkNodeState CR when update is…

pliurh:
… triggered

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -18,6 +18,7 @@ import (
 
 	"github.com/golang/glog"
 	drain "github.com/openshift/kubernetes-drain"
+	goerr "github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


pkg/daemon/daemon.go
```diff
@@ -281,9 +281,24 @@ func (dn *Daemon) nodeStateChangeHandler(old, new interface{}) {
...
+	if len(dn.LoadedPlugins) == 0 {
+		err := dn.loadVendorPlugins(newState)
+		if err != nil {
+			glog.Errorf("nodeStateChangeHandler(): failed to load vendor plugin: %v", err)
+			dn.exitCh <- err
+			return
+		}
+	}
+	// Get the latest NodeState
+	latestState, err := dn.client.SriovnetworkV1().SriovNetworkNodeStates(namespace).Get(dn.name, metav1.GetOptions{})
```
zshi-redhat, line 293: Can we guarantee `latestState` is the latest? I think there is a chance that daemon writer not updating node state immediately after policy is deleted.

pkg/daemon/daemon.go
```diff
@@ -281,9 +281,24 @@ func (dn *Daemon) nodeStateChangeHandler(old, new interface{}) {
 	}
 	reqReboot := false
 	reqDrain := false
+	if len(dn.LoadedPlugins) == 0 {
```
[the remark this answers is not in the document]
pliurh: This is to ensure that the plugins will only be loaded once. Normally, the plugins will be loaded when the daemon starts, and `nodeStateAddHandler` will invoke `nodeStateChangeHandler` for the first time.

pliurh:
/hold
I need to refactor this PR a bit.

pkg/daemon/daemon.go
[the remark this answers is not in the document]
pliurh: I removed the logic at the latest commit.

pkg/daemon/daemon.go
pliurh: All the CR change event will be put in the workqueue, and be processed one after another. So I add a logic to check if the spec of the newState is identical with the latestState, if not, the event will be skipped. It makes sure that only the event latestState will be processed.

pkg/daemon/daemon.go
```diff
@@ -270,8 +270,20 @@ func (dn *Daemon) nodeStateChangeHandler(old, new interface{}) {
 	var err error
 	newState := new.(*sriovnetworkv1.SriovNetworkNodeState)
 	oldState := old.(*sriovnetworkv1.SriovNetworkNodeState)
-	glog.V(2).Infof("nodeStateChangeHandler(): current generation is %d", newState.GetObjectMeta().GetGeneration())
-	if newState.GetObjectMeta().GetGeneration() == oldState.GetObjectMeta().GetGeneration() {
+	glog.V(2).Infof("nodeStateChangeHandler(): new generation is %d", newState.GetObjectMeta().GetGeneration())
+	// Get the latest NodeState
+	latestState, err := dn.client.SriovnetworkV1().SriovNetworkNodeStates(namespace).Get(dn.name, metav1.GetOptions{})
```
zshi-redhat, line 275: shall it use retry here to avoid transient failure (as the error handling is exiting the daemon)?

pliurh: fixed

PR #182, opened 2020-04-10
Add 154d to Supported Devices

mmirecki:


test/util/cluster/cluster.go
```diff
@@ -16,7 +16,7 @@ type EnabledNodes struct {
 
 var (
 	supportedDrivers = []string{"mlx5_core", "i40e", "ixgbe"}
-	supportedDevices = []string{"1583", "158b", "10fb", "1015", "1017"}
+	supportedDevices = []string{"1583", "158b", "10fb", "1015", "1017", "154d"}
```
zshi-redhat, line 19: There is hard-coded PF->VF mapping in SR-IOV [operator code](https://github.com/openshift/sriov-network-operator/blob/master/pkg/apis/sriovnetwork/v1/helper.go#L13) which is used by sriov config daemon to generate sriov device plugin config map. If we only change the supportedDevices in test code, it may not work as sriov config daemon may not be able to generate correct config map for sriov device plugin.

I don't think we want to add a new device ID for testing purpose in operator code because we intentionally filter the devices via hard-coded PF to VF mapping and only support selected devices.

@pliurh ^^

pliurh: Could you tell us why you add this device to your test?

mmirecki: One of our testing envs has this id, so it would just make it easier to use it for development.

PR #187, opened 2020-04-14
Use DeepDerivative to compare the the kube object content

pliurh:
`DeepEqual` compares all the fields in the objects, however, we want to only focus on the fields set by the controller. This PR will also fix the problem that the controller keeps reconciling, due to the `DeepEqual` always returns false.

### DIFF
--- a/pkg/apply/apply.go
+++ b/pkg/apply/apply.go
@@ -90,7 +90,7 @@ func ApplyObject(ctx context.Context, client k8sclient.Client, obj *uns.Unstruct
 	if err := MergeObjectForUpdate(existing, obj); err != nil {
 		return errors.Wrapf(err, "could not merge object %s with existing", objDesc)
 	}
-	if !equality.Semantic.DeepEqual(existing, obj) {
+	if !equality.Semantic.DeepDerivative(obj, existing) {
 		if err := client.Update(ctx, obj); err != nil {
 			return errors.Wrapf(err, "could not update object %s", objDesc)
 		} else {
--- a/pkg/apply/merge.go
+++ b/pkg/apply/merge.go
@@ -10,11 +10,6 @@ import (
 // This is to be able to do a a meaningful comparison in apply,
 // since objects created on runtime do not have these fields populated.
 func MergeMetadataForUpdate(current, updated *uns.Unstructured) error {
-	updated.SetCreationTimestamp(current.GetCreationTimestamp())
-	updated.SetSelfLink(current.GetSelfLink())
-	updated.SetGeneration(current.GetGeneration())
-	updated.SetUID(current.GetUID())
-	updated.SetResourceVersion(current.GetResourceVersion())
 
 	mergeAnnotations(current, updated)
 	mergeLabels(current, updated)
@@ -133,8 +128,9 @@ func mergeAnnotations(current, updated *uns.Unstructured) {
 	for k, v := range updatedAnnotations {
 		curAnnotations[k] = v
 	}
-
-	updated.SetAnnotations(curAnnotations)
+	if len(curAnnotations) > 1 {
+		updated.SetAnnotations(curAnnotations)
+	}
 }
 
 // mergeLabels copies over any labels from current to updated,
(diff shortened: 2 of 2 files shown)


pkg/apply/apply.go
```diff
@@ -90,7 +90,7 @@ func ApplyObject(ctx context.Context, client k8sclient.Client, obj *uns.Unstruct
 	if err := MergeObjectForUpdate(existing, obj); err != nil {
 		return errors.Wrapf(err, "could not merge object %s with existing", objDesc)
 	}
-	if !equality.Semantic.DeepEqual(existing, obj) {
+	if !equality.Semantic.DeepDerivative(obj, existing) {
```
zshi-redhat, line 93: So parameter order doesn't matter ?

pkg/apply/merge.go
```diff
@@ -133,8 +128,9 @@ func mergeAnnotations(current, updated *uns.Unstructured) {
 	for k, v := range updatedAnnotations {
 		curAnnotations[k] = v
 	}
-
-	updated.SetAnnotations(curAnnotations)
+	if len(curAnnotations) > 1 {
```
zshi-redhat, line 131: Will this block if from setting annotation to empty?

pkg/apply/merge.go
```diff
@@ -150,8 +146,9 @@ func mergeLabels(current, updated *uns.Unstructured) {
 	for k, v := range updatedLabels {
 		curLabels[k] = v
 	}
-
-	updated.SetLabels(curLabels)
+	if len(curLabels) > 1 {
```
zshi-redhat, line 149: Will this block if from setting labels to empty?

pkg/apply/merge.go
```diff
@@ -10,11 +10,6 @@ import (
 // This is to be able to do a a meaningful comparison in apply,
 // since objects created on runtime do not have these fields populated.
 func MergeMetadataForUpdate(current, updated *uns.Unstructured) error {
-	updated.SetCreationTimestamp(current.GetCreationTimestamp())
```
zshi-redhat, line 13: If all these are set to the same values as in `current` obj. how would it trigger the reconcile?

pkg/apply/apply.go
pliurh: DeepDerivative will ignore the fields which are not set in the first object. `obj` only contains the fields set by the controller intentionally.

pkg/apply/merge.go
pliurh: As abovementioned, we shall keep the obj only contains the fields that we really need.

pkg/apply/merge.go
pliurh: The original logic is a merging instead of replacing. So the controller can only add new labels or do nothing.

pkg/apply/merge.go
pliurh: If we keep using`DeepEqual `. There are many other object-specific fields that need to be set. , we need to create a very complicated logic to make sure that all necessary fields are copied before comparing.

zshi-redhat: If we change DeepEqual to DeepDerivative,  do we still need to remove the update here?

pliurh: We can still leave the update here, but it would not make any difference. As if those fields are unset, they will not be token account of.

zshi-redhat: My concern is If we remove the update, it may result in a CR without these fields.

pliurh: It will not happen, those metadata will be set by the server during the update.

zshi-redhat: The values set by server will be different from previous (current) ones or they are the same?

pliurh: They will be the same, except for `generation`, it will increase by every change. Actually those metadata are managed by the server, and cannot be set by a client. I think the old code here is just for fulfilling the comparison requirement of `DeepEqual`.

zshi-redhat: ok

PR #191, opened 2020-04-15
Raising a meaningful error message if iommu is not enabled in bios

### LINKED ISSUE #162: Raising a meaningful error message if iommu is not enabled in bios
When configuring a device to use the vfio driver, if IOMMU is not enabled in the bios of the machine, the resulting `SriovNetworkNodeState` will report a failure along the line of `write error: No such device` when trying to bind a device to the vfio driver.

My suggestion here is to check if iommu is enabled (by checking the content of /sys/class/iommu`, and provide a more descriptive error message, so that the admin can understand how to fix the error.

pliurh:
fix #162

### DIFF
--- a/pkg/utils/driver.go
+++ b/pkg/utils/driver.go
@@ -1,10 +1,12 @@
 package utils
 
 import (
-	"github.com/golang/glog"
+	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
+
+	"github.com/golang/glog"
 
 	dputils "github.com/intel/sriov-network-device-plugin/pkg/utils"
 	sriovnetworkv1 "github.com/openshift/sriov-network-operator/pkg/apis/sriovnetwork/v1"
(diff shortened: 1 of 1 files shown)


fedepaol:
Thanks!
/lgtm

pkg/utils/driver.go
```diff
@@ -55,6 +57,11 @@ func BindDpdkDriver(pciAddr, driver string) error {
 	err = ioutil.WriteFile(bindPath, []byte(pciAddr), os.ModeAppend)
 	if err != nil {
 		glog.Errorf("BindDpdkDriver(): fail to bind driver for device %s: %s", pciAddr, err)
+		_, err := os.Readlink(filepath.Join(sysBusPciDevices, pciAddr, "iommu_group"))
```
zshi-redhat, line 60: when driver bind fails, there might be two reasons:
1) iommu not enabled in BIOS.  (will this result in error when setting iommu kernel argument ?)
2) iommu is enabled in BIOS, iommu kernel argument is not configured.

Shall we distinguish the two with different error messages?

fedepaol: Isn't 2. covered by the operator itself?

zshi-redhat: There was an issue reported from customer-0 that vfio bind failed but VT-d is enabled in BIOS. We didn't find the root cause as there is no clear reproducer. The fix propsed for aforementioned issue is to at: https://github.com/openshift/sriov-network-operator/pull/179. But I don't know if there are other causes.

fedepaol: Understood, thanks

pliurh: After #179, the daemon will always reboot the node if the IOMMU is not enabled in Kernel. So if case 2, the daemon will reboot the host before trying to load the driver.

PR #199, opened 2020-04-24
Adding bond-cni install

mmirecki:


zshi-redhat:
@mmirecki are we still looking at adding bond-cni in sriov operator or it's done in CNO?

mmirecki:
> @mmirecki are we still looking at adding bond-cni in sriov operator or it's done in CNO?

We need it in sriov operator. CNO is not accepting any additional cni's unless absolutely neccessary for their key functionality (which the bond-cni is not).

zshi-redhat:
> > @mmirecki are we still looking at adding bond-cni in sriov operator or it's done in CNO?
> 
> We need it in sriov operator. CNO is not accepting any additional cni's unless absolutely neccessary for their key functionality (which the bond-cni is not).

@mmirecki  bond-cni is not just for sriov, but can also be used for host devices or any other network devices. If we put it in SR-IOV Operator, then user who wants to deploy bond will need to deploy SR-IOV Operator and configured SR-IOV policy first, which is not necessary for simple PF bonding.

mmirecki:
> > > @mmirecki are we still looking at adding bond-cni in sriov operator or it's done in CNO?
> > 
> > 
> > We need it in sriov operator. CNO is not accepting any additional cni's unless absolutely neccessary for their key functionality (which the bond-cni is not).
> 
> @mmirecki bond-cni is not just for sriov, but can also be used for host devices or any other network devices. If we put it in SR-IOV Operator, then user who wants to deploy bond will need to deploy SR-IOV Operator and configured SR-IOV policy first, which is not necessary for simple PF bonding. What is the current use case to address?

@zshi-redhat of course bond-cni can be used on top of any other cni. For this reason we tried to make it part of CNO, but this was rejected. Our scenario is mainly about using it with sriov, that's why we want to do the installation here. 
If the usecase of using it with simple nics will be accepted, we will need to move this to CNO.

zshi-redhat:
> > > > @mmirecki are we still looking at adding bond-cni in sriov operator or it's done in CNO?
> > > 
> > > 
> > > We need it in sriov operator. CNO is not accepting any additional cni's unless absolutely neccessary for their key functionality (which the bond-cni is not).
> > 
> > 
> > @mmirecki bond-cni is not just for sriov, but can also be used for host devices or any other network devices. If we put it in SR-IOV Operator, then user who wants to deploy bond will need to deploy SR-IOV Operator and configured SR-IOV policy first, which is not necessary for simple PF bonding. What is the current use case to address?
> 
> @zshi-redhat of course bond-cni can be used on top of any other cni. For this reason we tried to make it part of CNO, but this was rejected. Our scenario is mainly about using it with sriov, that's why we want to do the installation here.
> If the usecase of using it with simple nics will be accepted, we will need to move this to CNO.

@mmirecki   Ack.

bindata/manifests/daemon/daemonset.yaml
```diff
@@ -1,4 +1,88 @@
...
+kind: ConfigMap
+metadata:
+  name: cni-binary-copy-script
+  namespace: {{.Namespace}}
+  annotations:
+    kubernetes.io/description: |
+      This is a script used to copy CNI binaries based on host OS
+    release.openshift.io/version: "{{.ReleaseVersion}}"
+data:
+  cnibincopy.sh: |-
```
zshi-redhat, line 12: Can we add dockerfile/entrypoint.sh in intel/bond-cni repo and use that directly in bond-cni initContainer?

fepan:
> > > > @mmirecki are we still looking at adding bond-cni in sriov operator or it's done in CNO?
> > > 
> > > 
> > > We need it in sriov operator. CNO is not accepting any additional cni's unless absolutely neccessary for their key functionality (which the bond-cni is not).
> > 
> > 
> > @mmirecki bond-cni is not just for sriov, but can also be used for host devices or any other network devices. If we put it in SR-IOV Operator, then user who wants to deploy bond will need to deploy SR-IOV Operator and configured SR-IOV policy first, which is not necessary for simple PF bonding. What is the current use case to address?
> 
> @zshi-redhat of course bond-cni can be used on top of any other cni. For this reason we tried to make it part of CNO, but this was rejected. Our scenario is mainly about using it with sriov, that's why we want to do the installation here.
> If the usecase of using it with simple nics will be accepted, we will need to move this to CNO.

bond-cni should not be part of SR-IOV operator, the fact that it can be used with other interface types means we will likely get future requests to do so, and requiring users to install SR-IOV operator to deploy bond-cni is not the right path.

zshi-redhat:
> > > > > @mmirecki are we still looking at adding bond-cni in sriov operator or it's done in CNO?
> > > > 
> > > > 
> > > > We need it in sriov operator. CNO is not accepting any additional cni's unless absolutely neccessary for their key functionality (which the bond-cni is not).
> > > 
> > > 
> > > @mmirecki bond-cni is not just for sriov, but can also be used for host devices or any other network devices. If we put it in SR-IOV Operator, then user who wants to deploy bond will need to deploy SR-IOV Operator and configured SR-IOV policy first, which is not necessary for simple PF bonding. What is the current use case to address?
> > 
> > 
> > @zshi-redhat of course bond-cni can be used on top of any other cni. For this reason we tried to make it part of CNO, but this was rejected. Our scenario is mainly about using it with sriov, that's why we want to do the installation here.
> > If the usecase of using it with simple nics will be accepted, we will need to move this to CNO.
> 
> bond-cni should not be part of SR-IOV operator, the fact that it can be used with other interface types means we will likely get future requests to do so, and requiring users to install SR-IOV operator to deploy bond-cni is not the right path.

If we take all the possible pod-level bonding usages with bond-cni into consideration, I think the best approach is to include it in an Operator deployed in day 1, like CNO or maybe nmstate Operator. Whenever SR-IOV bonding is needed, it can just use the CNI instead of having to deploy another day 2 operator that may contain bond-cni.

Also I think there is timing issue on when SR-IOV bond, or PF bond or any other bonding will need to be addressed/supported. I only know SR-IOV bond was discussed and might be a request from CNF application at this point.

PR #200, opened 2020-04-28
Bug 1836212: Introduce a custom Dailer to close TCP connections when heartbeat fails

pliurh:
Force close connections on heartbeat failure, when heartbeat failed, we invoke closeAllConns and reestablish new connections. We use the writer's periodic Get() call as hearbeat singal, to detect whether the HTTP connection to the apiserver is still healthy.

### DIFF
--- a/pkg/daemon/writer.go
+++ b/pkg/daemon/writer.go
@@ -15,16 +15,18 @@ import (
 )
 
 type NodeStateStatusWriter struct {
-	client snclientset.Interface
-	node   string
-	status sriovnetworkv1.SriovNetworkNodeStateStatus
+	client             snclientset.Interface
+	node               string
+	status             sriovnetworkv1.SriovNetworkNodeStateStatus
+	OnHeartbeatFailure func()
 }
 
 // NewNodeStateStatusWriter Create a new NodeStateStatusWriter
-func NewNodeStateStatusWriter(c snclientset.Interface, n string) *NodeStateStatusWriter {
+func NewNodeStateStatusWriter(c snclientset.Interface, n string, f func()) *NodeStateStatusWriter {
 	return &NodeStateStatusWriter{
-		client: c,
-		node:   n,
+		client:             c,
+		node:               n,
+		OnHeartbeatFailure: f,
 	}
 }
 
@@ -38,7 +40,7 @@ func (writer *NodeStateStatusWriter) Run(stop <-chan struct{}, refresh <-chan Me
 		if err := writer.pollNicStatus(); err != nil {
 			glog.Errorf("Run(): first poll failed: %v", err)
 		}
-		setNodeStateStatus(writer.client, writer.node, writer.status, msg)
+		writer.setNodeStateStatus(msg)
 		return
 	}
 	for {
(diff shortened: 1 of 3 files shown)


pliurh:
This PR is inspired by https://github.com/kubernetes/kubernetes/pull/78016, and it can mitigate the problem caused by https://bugzilla.redhat.com/show_bug.cgi?id=1826769

pkg/daemon/writer.go
```diff
@@ -121,20 +123,24 @@ func setNodeStateStatus(client snclientset.Interface, nodeName string, status sr
...
 	err := wait.PollImmediate(10*time.Second, 5*time.Minute, func() (bool, error) {
-		n, lastErr = client.SriovnetworkV1().SriovNetworkNodeStates(namespace).Get(nodeName, metav1.GetOptions{})
+		n, lastErr = w.client.SriovnetworkV1().SriovNetworkNodeStates(namespace).Get(w.node, metav1.GetOptions{})
 		if lastErr == nil {
 			return true, nil
 		}
-		glog.Warningf("Failed to fetch node %s (%v); retrying...", nodeName, lastErr)
+		glog.Warningf("getNodeState(): Failed to fetch node state %s (%v); close all connections and retry...", w.node, lastErr)
+		// Use the Get() also as an client-go keepalive indicator for the TCP connection.
+		if w.OnHeartbeatFailure != nil {
```
fedepaol, line 136: I think this can't be null, but if so, wouldn't it better for this to explode instead of not working? Or, if you are on the defensive side, can you add a log here?

fedepaol:
Nice! 
So, if I got this right, the general idea here is that informers watch connection may be established but not working and we leverage the periodic `Get` to force them to disconnect (and automatically reconnect, I assume)

pkg/daemon/writer.go
pliurh: I think you're right.

pliurh:
> Nice!
> So, if I got this right, the general idea here is that informers watch connection may be established but not working and we leverage the periodic `Get` to force them to disconnect (and automatically reconnect, I assume)

Correct. This happens in my environment frequently. But I'm still waiting for QE's feedback with their environment. Normally, if the cluster network works as expected, the timeout won't happen. Have you experienced such an issue?

fedepaol:
> > Nice!
> > So, if I got this right, the general idea here is that informers watch connection may be established but not working and we leverage the periodic `Get` to force them to disconnect (and automatically reconnect, I assume)
> 
> Correct. This happens in my environment frequently. But I'm still waiting for QE's feedback with their environment. Normally, if the cluster network works as expected, the timeout won't happen. Have you experienced such an issue?

Nope, but I'll keep an eye on that

fedepaol:
> > > Nice!
> > > So, if I got this right, the general idea here is that informers watch connection may be established but not working and we leverage the periodic `Get` to force them to disconnect (and automatically reconnect, I assume)
> > 
> > 
> > Correct. This happens in my environment frequently. But I'm still waiting for QE's feedback with their environment. Normally, if the cluster network works as expected, the timeout won't happen. Have you experienced such an issue?
> 
> Nope, but I'll keep an eye on that

Update: we are definitely experiencing that.

PR #202, opened 2020-05-08
Bug 1830497: Use privileged SCC to replace runlevel

pliurh:


bindata/manifests/plugins/002-rbac.yaml
```diff
@@ -10,3 +10,61 @@ kind: ServiceAccount
 metadata:
   name: sriov-device-plugin
   namespace: {{.Namespace}}
+---
+apiVersion: rbac.authorization.k8s.io/v1
```
fedepaol, line 14: Could the role be the same for both the cni and the dp and then bind both accounts to the same role?

pliurh: Technically, yes. 
@zshi-redhat do you think DP and CNI pod will need different roles in the future?

zshi-redhat: @pliurh for the foreseeable future, I didn't see there is such need.

pliurh:
@fedepaol @zshi-redhat Now one role is defined for both CNI and DP SA. PTAL.

PR #204, opened 2020-05-09
Bug 1832728: Use operator webhook to prevent users from configuring unsupported de…

pliurh:
…vices

### DIFF
--- a/pkg/webhook/validate.go
+++ b/pkg/webhook/validate.go
@@ -19,6 +19,12 @@ const (
 	MlxMaxVFs  = 128
 )
 
+// SupportedModels holds the NIC models officially supported
+var SupportedModels = map[string]([]string){
+	IntelID:    []string{"158b"},
+	MellanoxID: []string{"1015", "1017"},
+}
+
 var (
 	nodesSelected     bool
 	interfaceSelected bool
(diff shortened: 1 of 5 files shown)


pkg/webhook/validate.go
```diff
@@ -57,6 +63,21 @@ func staticValidateSriovNetworkNodePolicy(cr *sriovnetworkv1.SriovNetworkNodePol
...
+			if (cr.Spec.NicSelector.Vendor == IntelID && !sriovnetworkv1.StringInArray(cr.Spec.NicSelector.DeviceID, SupportedModels[IntelID])) || (cr.Spec.NicSelector.Vendor == MellanoxID && !sriovnetworkv1.StringInArray(cr.Spec.NicSelector.DeviceID, SupportedModels[MellanoxID])) {
+				return false, fmt.Errorf("vendor/device %s/%s is not supported", cr.Spec.NicSelector.Vendor, cr.Spec.NicSelector.DeviceID)
+			}
+		}
+	} else if cr.Spec.NicSelector.DeviceID != "" {
+		if !sriovnetworkv1.StringInArray(cr.Spec.NicSelector.DeviceID, append(SupportedModels[IntelID], SupportedModels[MellanoxID]...)) {
+			return false, fmt.Errorf("device %s is not supported", cr.Spec.NicSelector.DeviceID)
+		}
+	}
+
```
zshi-redhat, line 80: I think we still allow use of other type of NICs if vendor/device are not explicitly specified. Shall we change the behavior as well?

pliurh: Users can still do that by disabling the operator webhook.

zshi-redhat: Right, but I think it's different issue when disabling webhook where we have no control over which IDs are configured.

pliurh: The implication is that, when users choose to disable the webhook, they shall know exactly what they are trying to do. Otherwise, it shall not be disabled.

Also, I agree to @dankenigsberg 's comments that we shall not have the validation logic at more than 1 place.

PR #210, opened 2020-05-12
Bug 1833603: Limit the range of logLevel and add detailed description

pliurh:


pliurh:
@zhaozhanqi PTAL

deploy/crds/sriovnetwork.openshift.io_sriovoperatorconfigs_crd.yaml
```diff
@@ -47,7 +47,11 @@ spec:
                 webhook shall be deployed
               type: boolean
             logLevel:
-              description: Flag to control the log verbose level of the operator
+              description: Flag to control the log verbose level of the operator.
```
zhaozhanqi, line 50: how about update 'operator' to 'config daemon' since for now the logLevel only control the sriov config daemon logs

pliurh: In the long term, we may also want to use this flag to configure the verbose level for the controller as well. So I prefer to keep this description more generic. @zshi what do you think?

zshi-redhat: Yes, this will be the single flag to control verbose log level for all operator managed components, including operator itself, I think we can leave it as generic.

zhaozhanqi: ok,then lgtm

rbbratta: it looks like the global OpenShift API defines `logLevel` as a string, should we use a string instead of integer?

https://github.com/openshift/api/blob/master/operator/v1/types.go#L55

```
type LogLevel string

var (
	// Normal is the default.  Normal, working log information, everything is fine, but helpful notices for auditing or common operations.  In kube, this is probably glog=2.
	Normal LogLevel = "Normal"

	// Debug is used when something went wrong.  Even common operations may be logged, and less helpful but more quantity of notices.  In kube, this is probably glog=4.
	Debug LogLevel = "Debug"

	// Trace is used when something went really badly and even more verbose logs are needed.  Logging every function call as part of a common operation, to tracing execution of a query.  In kube, this is probably glog=6.
	Trace LogLevel = "Trace"

	// TraceAll is used when something is broken at the level of API content/decoding.  It will dump complete body content.  If you turn this on in a production cluster
	// prepare from serious performance issues and massive amounts of logs.  In kube, this is probably glog=8.
	TraceAll LogLevel = "TraceAll"
)
```

PR #219, opened 2020-05-21
Initialize administrative mac for vfs

Mmduh-483:


zshi-redhat:
@Mmduh-483 could you elaborate why this is needed?

pkg/utils/utils.go
```diff
@@ -302,6 +302,23 @@ func getNetdevMTU(pciAddr string) int {
 	return mtu
 }
 
+func getNetMac(pciAddr string) string {
```
moshe010, line 305: can we use netlink to get the effective mac?

Mmduh-483:
> @Mmduh-483 could you elaborate why this is needed?

For mellnaox cards to use RoCE capabilities in ETH mode, it needs initialize the node guid and port guid, which is received from the administrative mac of the VFs (initially all 0s)

moshe010:
> > @Mmduh-483 could you elaborate why this is needed?
> 
> @zshi-redhat for mellnaox cards to use RoCE capabilities in ETH mode, it needs initialize the node guid and port guid, which is received from the administrative mac of the VFs (initially all 0s)

This is not accurate. RoCE is working but only without rdma_cm. The reason is the node_guid is zero  and rdma_cm used node_guid to id the connection in the host. We will fix it in the upstream kernel that RDMA traffic with rdma_cm will work without user set admin mac/unbind.bind the VF but until this will happen we want to do the suggested workaround in the SR-IOV operator.

pkg/utils/utils.go
Mmduh-483: changed

pliurh:
If we set the administrative mac for a VF, would that mac address be inherited when it's attached to a container? If yes, is it still possible for users to specify a custom mac address? And if those 2 mac addresses are different, will that be an problem when the spoof checking is enabled?

moshe010:
So the reason to set the VF mac is to allow to set to node_guid for RDMA traffic (with RDMA_CM) will work in the start of the world. (Like I said this is a workaround for driver issue).

In case you use spoof check off there is no problem. 
I case you use spoofcheck on you anyway need to specify the mac address and the SR-IVO cni will put the administrative mac == effective mac . the node guid will stay the same random which is ok because  mac address and guid are orthogonal

pkg/plugins/mellanox/mellanox_plugin.go
```diff
@@ -111,6 +114,35 @@ func (p *MellanoxPlugin) OnNodeStateChange(old, new *sriovnetworkv1.SriovNetwork
 
 	if needReboot {
 		needDrain = true
+	} else {
```
zshi-redhat, line 117: If node gets reboot for various reason, will this be re-configured if node specific policy is not changed?

zshi-redhat: Should we set the admin MAC address right after VFs are provisioned in config daemon?

Mmduh-483: The admin mac should be set after the VFs created, since the VFs admin mac will be all zeros when creating the VFs.

Mmduh-483: In case of reboot, the VFs will be recreated after reboot

moshe010: @zshi-redhat , you mean that this change will not work is the node was just reboot because OnNodeStateChange won't be called? What other ways we can do it?

zshi-redhat: @Mmduh-483 @moshe010 yes, my concern is if there is no policy change for a specific node (meaning OnNodeStateChange won't be called), when node gets rebooted randomly. Then admin MAC won't be set. Even if it's called, it will return immediately if SriovNetworkNodeState is not newly generated. https://github.com/openshift/sriov-network-operator/blob/master/pkg/daemon/daemon.go#L309

I remembered @pliurh mentioned that `OnNodeStateAdd` will be called every time when node gets rebooted, this is something that guaranteed by the informer mechanism.

It case the latter (informer always triggers OnNodeStateAdd) after rebooting is true, this change may work.

WRT changing MAC address when provisioning VFs, I'm inclined to setting the admin MAC address for all NICs in generic plugin regardless which vendor/device IDs it may have so that user can expect consistent behavior from Operator.

moshe010: I see, @Mmduh-483, Please check the reboot case and see if we need to move the logic to OnNodeStateAdd

Mmduh-483: @zshi-redhat `OnNodeStateChange` wasn't called after the reboot

moshe010: > I see, @Mmduh-483, Please check the reboot case and see if we need to move the logic to OnNodeStateAdd

moshe010: > @Mmduh-483 @moshe010 yes, my concern is if there is no policy change for a specific node (meaning OnNodeStateChange won't be called), when node gets rebooted randomly. Then admin MAC won't be set. Even if it's called, it will return immediately if SriovNetworkNodeState is not newly generated. https://github.com/openshift/sriov-network-operator/blob/master/pkg/daemon/daemon.go#L309
> 
> I remembered @pliurh mentioned that `OnNodeStateAdd` will be called every time when node gets rebooted, this is something that guaranteed by the informer mechanism.
> 
> It case the latter (informer always triggers OnNodeStateAdd) after rebooting is true, this change may work.
> 
> WRT changing MAC address when provisioning VFs, I'm inclined to setting the admin MAC address for all NICs in generic plugin regardless which vendor/device IDs it may have so that user can expect consistent behavior from Operator.

So this is a workaround for Mellanox driver and we should remove it one the kernel will be fix. So I don't thing we should do it in the generic  plugins?

zshi-redhat: I actually considered to make this as generic for all NICs even without this PR or issue of RDMA. One of the reasons is to provide consistent user experience when configuring SR-IOV on various vendor NICs.
For example, the default value for spoofChk is different with Mellanox and Intel NICs, and the behavior of traffic spoofing is also different when `spoofChk=on` && `admin-MAC=00:00:00:00:00:00`.

I'd expect there might be other differences going forward, so adding this change in generic plugin for all VFs allows us to reset VFs to the same initial state. It may later be extended to reset other VF attributes as well.

moshe010: @zshi-redhat, I see Ok we can do it, but first we need to fix the reboot case which is not working can you comment on why OnNodeStateChange or OnNodeStateAdd wasn't called after the reboot, Shouldn't there be a periodic sync?

zshi-redhat: > @zshi-redhat, I see Ok we can do it, but first we need to fix the reboot case which is not working can you comment on why OnNodeStateChange or OnNodeStateAdd wasn't called after the reboot, Shouldn't there be a periodic sync?

@moshe010 @Mmduh-483  yes, there is a periodic trigger of the OnNodeStateChange call, the default rsync duration is set to 15 seconds. If there is no policy being applied, you shall be able to see below log message in roughly every 15 seconds interval (it may not follow the exact 15 seconds time due to various interruptions ):

`2020-05-26T13:58:39.684947022Z I0526 13:58:39.684923    8803 daemon.go:285] nodeStateChangeHandler(): new generation is 3`

The number of new generation here is 3, it varies depending on the generation of applied policies.

Mmduh-483: I keep getting the message, but still `OnNodeStateChange` or `OnNodeStateAdd` not called yet.
```
I0528 06:04:04.905824    4216 utils.go:285] getNetdevMTU(): get MTU for device 0000:81:00.0
I0528 06:04:04.905880    4216 utils.go:280] tryGetInterfaceName(): name is enp129s0
I0528 06:04:04.905962    4216 utils.go:280] tryGetInterfaceName(): name is enp129s0
I0528 06:04:04.965595    4216 daemon.go:310] nodeStateChangeHandler(): Interface not changed
I0528 06:04:04.965641    4216 daemon.go:285] nodeStateChangeHandler(): new generation is 9
I0528 06:04:04.992940    4216 daemon.go:310] nodeStateChangeHandler(): Interface not changed
I0528 06:04:04.992988    4216 daemon.go:285] nodeStateChangeHandler(): new generation is 9
I0528 06:04:04.993116    4216 writer.go:115] setNodeStateStatus(): syncStatus: Succeeded, lastSyncError: 
I0528 06:04:04.998391    4216 daemon.go:310] nodeStateChangeHandler(): Interface not changed
I0528 06:04:04.998424    4216 daemon.go:285] nodeStateChangeHandler(): new generation is 9
I0528 06:04:05.004178    4216 daemon.go:310] nodeStateChangeHandler(): Interface not changed
I0528 06:04:05.004207    4216 daemon.go:285] nodeStateChangeHandler(): new generation is 9
```

pliurh: @Mmduh-483 Can you check if the `generation` in the SirovNetworkNodeState CR is 9 or not?

Mmduh-483: @pliurh I had to re-deploy the setup, Here is the logs
```
I0531 09:09:31.119887    5001 utils.go:280] tryGetInterfaceName(): name is enp5s0f0
I0531 09:09:31.120007    5001 utils.go:285] getNetdevMTU(): get MTU for device 0000:05:00.1
I0531 09:09:31.120054    5001 utils.go:280] tryGetInterfaceName(): name is enp5s0f1
I0531 09:09:31.120125    5001 utils.go:280] tryGetInterfaceName(): name is enp5s0f1
I0531 09:09:31.125976    5001 writer.go:115] setNodeStateStatus(): syncStatus: Succeeded, lastSyncError: 
I0531 09:09:32.393658    5001 daemon.go:285] nodeStateChangeHandler(): new generation is 4
I0531 09:09:32.400017    5001 daemon.go:310] nodeStateChangeHandler(): Interface not changed
I0531 09:09:47.393900    5001 daemon.go:285] nodeStateChangeHandler(): new generation is 4
I0531 09:09:47.399755    5001 daemon.go:310] nodeStateChangeHandler(): Interface not changed
[root@r-dcs07 sriov-network-operator]# kubectl describe sriovnetworknodestates.sriovnetwork.openshift.io -n sriov-network-operator r-dcs5-07
Name:         r-dcs5-07
Namespace:    sriov-network-operator
Labels:       <none>
Annotations:  <none>
API Version:  sriovnetwork.openshift.io/v1
Kind:         SriovNetworkNodeState
Metadata:
  Creation Timestamp:  2020-05-31T08:39:35Z
  Generation:          4
```
The generation is 4 but still there is no periodic call after reboot

Mmduh-483: @pliurh After reboot, The operator calls the `OnNodeStateAdd` once, but the status doesn't have the VFs, I added a log to print the VFs but the list `ifaceStatus.VFs` was empty.

The the call for `OnNodeStateAdd` the operator started discovering the VFs and PFs:

Here is the logs
```
I0601 07:09:07.214975    5185 mellanox_plugin.go:283] mellanox-plugin parseMstconfigOutput(): Attributes [NUM_OF_VFS SRIOV_EN LINK_TYPE_P1]
I0601 07:09:07.243357    5185 mellanox_plugin.go:250] mellanox-plugin mlnxNicFromMap() map[LINK_TYPE_P1:ETH(2) NUM_OF_VFS:13 SRIOV_EN:True(1)], true
I0601 07:09:07.243437    5185 mellanox_plugin.go:250] mellanox-plugin mlnxNicFromMap() map[LINK_TYPE_P1:ETH(2) NUM_OF_VFS:13 SRIOV_EN:True(1)], true
I0601 07:09:07.382374    5185 mellanox_plugin.go:120] MMMMMMMMMMMM vfs []
I0601 07:09:07.382408    5185 mellanox_plugin.go:140] MMMMMMMMMMMM vfs to rebind []
I0601 07:09:07.382421    5185 mellanox_plugin.go:145] mellanox-plugin needDrain false needReboot false
I0601 07:09:07.479264    5185 intel_plugin.go:37] intel-plugin OnNodeStateAdd()
I0601 07:09:07.628059    5185 generic_plugin.go:55] generic-plugin OnNodeStateAdd()
I0601 07:09:07.657274    5185 daemon.go:200] nodeStateAddHandler(): reqDrain true, reqReboot false
I0601 07:09:07.657329    5185 daemon.go:203] nodeStateAddHandler(): drain node
I0601 07:09:07.657344    5185 daemon.go:507] drainNode(): Update prepared
I0601 07:09:24.047041    5185 daemon.go:532] drainNode(): Check if any other node is draining
I0601 07:09:24.047098    5185 daemon.go:540] drainNode(): No other node is draining, stop watching
I0601 07:09:24.047396    5185 daemon.go:474] annotateNode(): Annotate node r-dcs5-07 with: Draining
I0601 07:09:24.080254    5185 daemon.go:565] drainNode(): Start draining
I0601 07:09:24.088676    5185 daemon.go:470] cordoned node "r-dcs5-07"
I0601 07:09:24.146152    5185 daemon.go:466] ignoring DaemonSet-managed pods: kube-multus-ds-amd64-hkk2z, kube-proxy-7hjjv, sriov-cni-n6dk9, sriov-device-plugin-frp9d, sriov-network-config-daemon-2hmj8
I0601 07:09:24.146189    5185 daemon.go:470] drained node "r-dcs5-07"
I0601 07:09:24.146213    5185 daemon.go:586] drainNode(): drain complete
I0601 07:09:24.146224    5185 daemon.go:474] annotateNode(): Annotate node r-dcs5-07 with: Idle
I0601 07:09:24.162021    5185 mellanox_plugin.go:151] mellanox-plugin Apply()
I0601 07:09:24.162097    5185 mellanox_plugin.go:168] mellanox-plugin configFW()
I0601 07:09:24.162230    5185 intel_plugin.go:49] intel-plugin Apply()
I0601 07:09:24.172715    5185 generic_plugin.go:105] generic-plugin Apply(): desiredState={4239549 [{0000:02:00.0 13 9000 enp2s0f0 [{mlxnics40 netdevice 0-12 policy4-0}]}]}
I0601 07:09:24.180406    5185 utils.go:133] needUpdate(): MTU needs update, desired=9000, current=1500
I0601 07:09:24.198459    5185 utils.go:170] configSriovDevice(): config interface 0000:02:00.0 with &{0000:02:00.0 13 9000 enp2s0f0 [{mlxnics40 netdevice 0-12 policy4-0}]}
I0601 07:09:24.198523    5185 utils.go:235] setSriovNumVfs(): set NumVfs for device 0000:02:00.0
I0601 07:09:28.730377    5185 writer.go:60] Run(): period refresh
I0601 07:09:28.730437    5185 writer.go:70] pollNicStatus()
I0601 07:09:28.730449    5185 utils.go:35] DiscoverSriovDevices
I0601 07:09:28.796875    5185 utils.go:285] getNetdevMTU(): get MTU for device 0000:02:00.0
I0601 07:09:28.797042    5185 utils.go:280] tryGetInterfaceName(): name is enp2s0f0
I0601 07:09:28.797488    5185 utils.go:280] tryGetInterfaceName(): name is enp2s0f0
I0601 07:09:29.481326    5185 utils.go:252] setNetdevMTU(): set MTU for device 0000:02:00.0
I0601 07:09:29.484361    5185 utils.go:280] tryGetInterfaceName(): name is enp2s0f2
I0601 07:09:29.484387    5185 utils.go:285] getNetdevMTU(): get MTU for device 0000:02:00.2
I0601 07:09:29.484451    5185 utils.go:280] tryGetInterfaceName(): name is enp2s0f2
I0601 07:09:29.484835    5185 utils.go:280] tryGetInterfaceName(): name is enp2s0f3
I0601 07:09:29.484848    5185 utils.go:285] getNetdevMTU(): get MTU for device 0000:02:00.3
I0601 07:09:29.484892    5185 utils.go:280] tryGetInterfaceName(): name is enp2s0f3
I0601 07:09:29.485820    5185 utils.go:280] tryGetInterfaceName(): name is enp2s1f4
I0601 07:09:29.485833    5185 utils.go:285] getNetdevMTU(): get MTU for device 0000:02:01.4
I0601 07:09:29.485897    5185 utils.go:280] tryGetInterfaceName(): name is enp2s1f4
I0601 07:09:29.486916    5185 utils.go:280] tryGetInterfaceName(): name is enp2s1f5
I0601 07:09:29.486928    5185 utils.go:285] getNetdevMTU(): get MTU for device 0000:02:01.5
```

pliurh: @Mmduh-483 
The VFs will only be recreated after reboot. So when configure daemon is up after a node reboot, the VFs are been created before `OnNodeStateAdd` is invoked.

I suggest we talk on Intel's slack.

Mmduh-483:
@pliurh @zshi-redhat 
Updated the PR to assign the mac address same as the effective mac when the VFs are created

pkg/utils/utils.go
```diff
@@ -196,13 +197,30 @@ func configSriovDevice(iface *sriovnetworkv1.Interface, ifaceStatus *sriovnetwor
 		if err != nil {
 			glog.Warningf("configSriovDevice(): unable to parse VFs for device %+v %q", iface.PciAddress, err)
 		}
+		ifaceLink, err := netlink.LinkByName(iface.Name)
```
zshi-redhat, line 200: @Mmduh-483  
Would you please create a separate function for changing admin MAC address and put it right after `setSriovNumVfs` is called? With this, it will only be called once when VFs are provisioned and it's guaranteed that all the VFs will be kernel netdevice interfaces (rather than bind to vfio-pci).

Mmduh-483: Changed

zshi-redhat:
| /test e2e-aws
-- | --

pkg/utils/utils.go
```diff
@@ -373,3 +377,37 @@ func Chroot(path string) (func() error, error) {
...
+		vfName := tryGetInterfaceName(addr)
+		vfLink, err := netlink.LinkByName(vfName)
+		if err != nil {
+			glog.Errorf("setVfsAdminMac(): unable to get VF link for device %+v %q", iface, err)
+			return err
+		}
+		if err := netlink.LinkSetVfHardwareAddr(pfLink, vfID, vfLink.Attrs().HardwareAddr); err != nil {
+			return err
+		}
+		if err = BindDefaultDriver(addr); err != nil {
```
zshi-redhat, line 407: @Mmduh-483 I just noticed this line, why do we need to bind default driver here? I assume when the VFs are first provisioned, they all bind to its corresponding vf drivers.

moshe010:
@zshi-redhat, the unbind and bind will set the node_guid. There is know other way to do it today.
 Now I remember that this is way we implmeneted it only for mellanox plugin

pkg/utils/utils.go
moshe010: So set admin mac and rebind the VF driver set the node_guid. This is the issue we wanted to fix in the first place. 
We need node_guid to so that it will work with rdma_cm

zshi-redhat: This line only bind default driver, not unbind, does it actually work with rdma_cm ?

moshe010: good point @Mmduh-483  is retesting it again. It maybe a bug

zshi-redhat: once the unbind/bind is verified, we can see how to move the unbind/bind logic in mlx plugin, and leave the admin MAC settings in generic plugin.

Mmduh-483:
@zshi-redhat @moshe010 
The generic plugin which create the VFs is the last to be called, so there is no other plugin that can do the unbind/bind for the VFs  because they are not created yet, and there is no periodic call for `OnNodeStateChange`, and there is no function to be called after `Apply`.

```
I0604 13:02:37.879931   17734 intel_plugin.go:49] intel-plugin Apply()
I0604 13:02:37.880008   17734 mellanox_plugin.go:166] mellanox-plugin Apply()
I0604 13:02:37.880371   17734 generic_plugin.go:105] generic-plugin Apply(): desiredState={4930843 [{0000:02:00.0 7 9000 enp2s0f0 [{mlxnics57 netdevice 0-6 policy57-0}]}]}
I0604 13:02:37.880458   17734 generic_plugin.go:115] generic-plugin Apply(): lastStat={4930317 []}
I0604 13:02:37.880542   17734 utils.go:134] needUpdate(): MTU needs update, desired=9000, current=1500
```

PR #220, opened 2020-05-21
Get more info about PFs and VFs

Mmduh-483:
PF: Mac address and linkSpeed
VF: Mac address and name

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -79,6 +79,8 @@ func DiscoverSriovDevices() ([]sriovnetworkv1.InterfaceExt, error) {
 		}
 		if name := tryGetInterfaceName(device.Address); name != "" {
 			iface.Name = name
+			iface.Mac = getNetDevMac(name)
+			iface.LinkSpeed = getNetDevLinkSpeed(name)
 		}
 
 		if dputils.IsSriovPF(device.Address) {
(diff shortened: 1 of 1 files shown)


pkg/utils/utils.go
```diff
@@ -302,6 +304,30 @@ func getNetdevMTU(pciAddr string) int {
 	return mtu
 }
 
+func getNetDevMac(ifaceName string) string {
```
pliurh, line 307: Using the pci address is more reliable, as sometimes the interface is available.

moshe010: how can you get MAC address with pci address?

pliurh: You can also find the mac address from the sysfs. for example, /sys/bus/pci/devices/0000:01:00.0/net/eno1/address. However it would still require the interface name. So let keep it as it is.

pkg/utils/utils.go
```diff
@@ -302,6 +304,30 @@ func getNetdevMTU(pciAddr string) int {
...
+func getNetDevLinkSpeed(ifaceName string) string {
+	glog.V(2).Infof("getNetDevLinkSpeed(): get LinkSpeed for device %s", ifaceName)
+	speedFilePath := filepath.Join(sysClassNet, ifaceName, "speed")
+	data, err := ioutil.ReadFile(speedFilePath)
+	if err != nil {
+		glog.Warningf("getNetDevLinkSpeed(): fail to read Link Speed file %s", speedFilePath)
+		return ""
+	}
+
+	return strings.TrimSpace(string(data))
```
zshi-redhat, line 328: nit: if we could append speed unit to the string, for example, it returns 10000, and becomes 10000 Mb/s after appending the unit.

Mmduh-483: added

PR #222, opened 2020-05-26
Bug 1840104: Make the daemon more resilient to failures.

fedepaol:
When the get nodes in uncordon fails, the daemon exits and the node remains uncordoned.

When the get of the device plugin pod fails, the daemon does not restart it and the change requested does not get applied.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -30,6 +30,7 @@ import (
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/cache"
+
 	// "k8s.io/client-go/kubernetes/scheme"
 	sriovnetworkv1 "github.com/openshift/sriov-network-operator/pkg/apis/sriovnetwork/v1"
 	snclientset "github.com/openshift/sriov-network-operator/pkg/client/clientset/versioned"
@@ -248,12 +249,7 @@ func (dn *Daemon) nodeStateAddHandler(obj interface{}) {
 
 func (dn *Daemon) uncordon(name string) {
 	glog.Info("uncordon(): uncordon node")
-	node, err := dn.kubeClient.CoreV1().Nodes().Get(name, metav1.GetOptions{})
-	if err != nil {
-		glog.Errorf("uncordon(): failed to get node: %v", err)
-		dn.exitCh <- err
-		return
-	}
+
 	backoff := wait.Backoff{
 		Steps:    5,
 		Duration: 10 * time.Second,
@@ -262,7 +258,14 @@ func (dn *Daemon) uncordon(name string) {
 	var lastErr error
 
 	if err := wait.ExponentialBackoff(backoff, func() (bool, error) {
-		err := drain.Uncordon(dn.kubeClient.CoreV1().Nodes(), node, nil)
+		node, err := dn.kubeClient.CoreV1().Nodes().Get(name, metav1.GetOptions{})
+		if err != nil {
+			glog.Infof("uncordon(): failed to get node: %v", err)
+			lastErr = err
+			return false, nil
+		}
+
+		err = drain.Uncordon(dn.kubeClient.CoreV1().Nodes(), node, nil)
 		if err == nil {
 			return true, nil
 		}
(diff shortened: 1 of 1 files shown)


fedepaol:
/hold
needs validation

fedepaol:
/hold cancel

Validated it with the first set of tests of conformance suite

pkg/daemon/daemon.go
```diff
@@ -379,39 +382,68 @@ func (dn *Daemon) restartDevicePluginPod() error {
...
+		}
+
+		if len(pods.Items) == 0 {
+			glog.Info("restartDevicePluginPod(): device plugin pod exited")
+			return true, nil
+		}
+		podName = pods.Items[0].Name
+		return true, nil
+	}, dn.stopCh); err != nil {
+		glog.Errorf("restartDevicePluginPod(): failed to wait for pod deletion: %v", err)
```
pliurh, line 410: This line logs the same message as line 415 and 441. I think it would be better to give different log message at different places.

pkg/daemon/daemon.go
```diff
@@ -379,39 +382,68 @@ func (dn *Daemon) restartDevicePluginPod() error {
...
+			glog.Info("restartDevicePluginPod(): device plugin pod exited")
+			return true, nil
+		}
+
+		if err != nil {
+			glog.Warningf("restartDevicePluginPod(): Failed to list device plugin pod: %s, retrying", err)
+			return false, nil
+		}
+
+		if len(pods.Items) == 0 {
```
zshi-redhat, line 403: In which case will the condition be met?

pkg/daemon/daemon.go
```diff
@@ -379,39 +382,68 @@ func (dn *Daemon) restartDevicePluginPod() error {
...
+	if podName == "" {
+		glog.Info("restartDevicePluginPod(): device plugin pod was not there")
 		return nil
 	}
-	if len(pods.Items) > 0 {
-		glog.V(2).Infof("restartDevicePluginPod(): Found device plugin pod %s", pods.Items[0].GetName())
-		err = dn.kubeClient.CoreV1().Pods(namespace).Delete(pods.Items[0].GetName(), &metav1.DeleteOptions{})
+
+	if err := wait.PollImmediateUntil(3*time.Second, func() (bool, error) {
+		glog.V(2).Infof("restartDevicePluginPod(): Found device plugin pod %s, deleting it", podName)
```
zshi-redhat, line 420: what happens if `len(pods.Items) == 0`? Will it cause config daemon exit?

zshi-redhat:
So until we get to the root cause, are we going to consider wrapping every API call into wait loop?
There are functions that access APIs such as `annotateNode`, `drainNode` which only try once.

pkg/daemon/daemon.go
fedepaol: I was checking for len > 0 in the previous version, so I assumed it was possible to have a len == 0.
Not sure about the behaviour of `List` in case of no matches.

pkg/daemon/daemon.go
fedepaol: You mean if there are no pods when we reach the delete? I am not sure if deleting a pod that does not exist returns an error. In that case the behaviour would be even worse (infinite loop).

This should not happen given the current flow (apart from external factors), but let me check if the error is notfound

pkg/daemon/daemon.go
fedepaol: ok

pkg/daemon/daemon.go
zshi-redhat: In previous version, it checks `pods.items >0` before deleting the pod.
With current change, there is no conditional check. I'm trying to understand why it's removed.

fedepaol: Ah, that's because if `len (pods.Items) == 0` we don't find a pod to delete and we exit in line 414.

pkg/daemon/daemon.go
zshi-redhat: In the case of len==0, this snippet returns true and it continues the next wait which delete the pod.
If the len is 0, how can it be deleted.

fedepaol: Nope, if len==0, podName does not get assigned and the function does not go up to the delete because it exits before.

Let me add a podToDelete bool to make it more clear.
If you can't read it, it's pretty clear the code wasn't well written.

pkg/daemon/daemon.go
zshi-redhat: ack, I missed the pfName checking. sorry for the noise.

pkg/daemon/daemon.go
zshi-redhat: ack

fedepaol:
/hold
Want to give this another run

PR #227, opened 2020-05-28
Use host netowrk for operator instead of CNI network

Mmduh-483:


zshi-redhat:
@Mmduh-483 Is there any issue with not using hostNetwork?

Mmduh-483:
> @Mmduh-483 Is there any issue when hostNetwork is not used?

Currently the operator is using the network CNI for the main interface, which sometimes may fail to deploy it because the cni failed, or because the cni interface can't reach the k8s api, I faced the second issue many times.

pliurh:
> > @Mmduh-483 Is there any issue when hostNetwork is not used?
> 
> Currently the operator is using the network CNI for the main interface, which sometimes may fail to deploy it because the cni failed, or because the cni interface can't reach the k8s api, I faced these two issues many times.

I don't think we should use hostNetwork here, as this is the controller of the operator which doesn't require `hostNetwork` to do its job. If the CNI interface doesn't work, that is a problem of that CNI, not the operator.

Mmduh-483:
> I don't think we should use hostNetwork here, as this is the controller of the operator which doesn't require `hostNetwork` to do its job. If the CNI interface doesn't work, that is a problem of that CNI, not the operator.

Since it is the controller it would serve better for it to make it not depending on a CNI network same like  the daemon is using `hostNetwork`

Mmduh-483:
@pliurh friendly reminder

pliurh:
> @pliurh friendly reminder

@Mmduh-483 I don't see the necessity to make such change for the operator. Normally the cluster network shall be reliable, and a pod shall require privilege as little as possible for security consideration.

Mmduh-483:
> > @pliurh friendly reminder
> 
> @Mmduh-483 I don't see the necessity to make such change for the operator. Normally the cluster network shall be reliable, and a pod shall require privilege as little as possible for security consideration.

PR #228, opened 2020-06-01
install sriov-infiniband-cni along with sriov-cni

zshi-redhat:


zshi-redhat:
/hold
hold until the image is built by ART.

bindata/manifests/plugins/sriov-cni.yaml
```diff
@@ -27,6 +27,12 @@ spec:
       tolerations:
       - operator: Exists
       serviceAccountName: sriov-cni
+      initContainers:
```
pliurh, line 30: Why using initContainter for ib-cni? Cann't we use the same approach as the sriov-cni?

zshi-redhat: To save system resource, With initContainer, we just use one daemonset for all CNI binary copy, this is the same as how Multus related CNI binaries are copied in cluster network operator.

PR #229, opened 2020-06-02
Allow to disable node reboot

mmirecki:
If the SriovOperatorConfig contains disableReboot=true the sriov-network-config-daemon will not reboot the nodes but instead just log this and exit.
(replaces PR 116 which was orphaned by github)


pkg/daemon/daemon.go
```diff
@@ -51,6 +53,8 @@ type Daemon struct {
 	client snclientset.Interface
 	// kubeClient allows interaction with Kubernetes, including the node we are running on.
 	kubeClient *kubernetes.Clientset
+	// kubeClient allows interaction with Kubernetes crd,s
+	dynamicclient dynamic.Interface
```
pliurh, line 57: There has already been an informer who monitors the SriovOperatorConfig CR. I don't think we need an extra client for the same task.

zshi-redhat:
@mmirecki I think this is only for debugging purpose, right? is there any other use cases ? for example: to coordinate multiple reboots into one, by disabling sriov reboot?

I'm aware of one use case that when master & worker are on the same nodes, node draining may fail as some of the pods cannot be evicted. like etcd. so it might be good if there is another option for disabling node draining.

mmirecki:
> @mmirecki I think this is only for debugging purpose, right? is there any other use cases ? for example: to coordinate multiple reboots into one, by disabling sriov reboot?
> 
> I'm aware of one use case that when master & worker are on the same nodes, node draining may fail as some of the pods cannot be evicted. like etcd. so it might be good if there is another option for disabling node draining.

Yes, this is meant for the QA environments.

mmirecki:
> > @mmirecki I think this is only for debugging purpose, right? is there any other use cases ? for example: to coordinate multiple reboots into one, by disabling sriov reboot?
> > I'm aware of one use case that when master & worker are on the same nodes, node draining may fail as some of the pods cannot be evicted. like etcd. so it might be good if there is another option for disabling node draining.
> 
> Yes, this is meant for the QA environments.

I talked to our QA's, and they say this feature is obsolete.
Closing PR

mmirecki:
Feature obsolete.
QA does not need this anymore.

PR #231, opened 2020-06-03
Share a SriovNetworkNodePolicy in selected tests

mmirecki:
The conformance tests take a very long time to execute. Most of this time is taken by environment reconfiguration and waiting for the env to stabilize, which usually takes between 200 to 400 seconds.
Some tests do not require an environment reconfiguration, but can instead be run on a single shared SriovNetworkNodePolicy. This allows these test to execute in an average of 20 seconds.

This patch groups the tests which can use a shared SriovNetworkNodePolicy together.


test/util/namespaces/namespaces.go
```diff
@@ -46,29 +46,35 @@ func Create(namespace string, cs *testclient.ClientSet) error {
 	return err
 }
 
-// Clean cleans all dangling objects from the given namespace.
-func Clean(operatorNamespace, namespace string, cs *testclient.ClientSet) error {
+func namespaceExists(namespace string, cs *testclient.ClientSet) bool {
 	_, err := cs.Namespaces().Get(namespace, metav1.GetOptions{})
-	if err != nil && k8serrors.IsNotFound(err) {
+	return err == nil || k8serrors.IsNotFound(err)
```
fedepaol, line 51: When k8serrors.IsNotFound(err) this should return false, because the namespace does not exists

test/conformance/tests/sriov_operator.go
```diff
@@ -51,273 +51,13 @@ var _ = Describe("[sriov] operator", func() {
...
-				err = clients.Create(context.Background(), secondConfig)
-				Expect(err).To(HaveOccurred())
-			})*/
+	Describe("Shared SriovNetworkNodePolicy", func() {
+		BeforeEach(func() {
+			err := namespaces.CleanPods(namespaces.Test, clients)
+			Expect(err).ToNot(HaveOccurred())
+			err = namespaces.CleanNetworks(operatorNamespace, clients)
+			Expect(err).ToNot(HaveOccurred())
+			waitForSRIOVStable()
```
fedepaol, line 60: Do we need this if we don't change the configuration?

test/conformance/tests/sriov_operator.go
```diff
@@ -379,15 +119,14 @@ var _ = Describe("[sriov] operator", func() {
 				validationFunction([]string{sriovNetwork.Name}, checkFunc)
 			}
 
-			BeforeEach(func() {
-				var err error
+			execute.BeforeAll(func() {
```
fedepaol, line 122: Can we put this on the outer level? It will be more clear that the same policy is shared across all the tests

test/conformance/tests/sriov_operator.go
```diff
@@ -681,292 +423,566 @@ var _ = Describe("[sriov] operator", func() {
...
+							},
+						}), ContainElement(corev1.DownwardAPIVolumeFile{
+							Path: "annotations",
+							FieldRef: &corev1.ObjectFieldSelector{
+								APIVersion: "v1",
+								FieldPath:  "metadata.annotations",
+							},
+						})))
+				})
+				Context("Multiple sriov device and attachment", func() {
```
fedepaol, line 793: This is nested in `Context("Resource Injector", func() { ` but I think it should be on the same level (it was already wrong before). Also, I think this does not depend on a specific policy either.

test/conformance/tests/sriov_operator.go
```diff
@@ -681,292 +423,566 @@ var _ = Describe("[sriov] operator", func() {
...
+				})
+				Context("IPv6 configured secondary interfaces on pods", func() {
+					// 25874
+					It("should be able to ping each other", func() {
+						resourceName := "sriovnic"
+						ipv6NetworkName := "test-ipv6network"
+						testNode := sriovInfos.Nodes[0]
+						sriovDevice, err := sriovInfos.FindOneSriovDevice(testNode)
+						Expect(err).ToNot(HaveOccurred())
+						createSriovPolicy(sriovDevice.Name, testNode, 5, resourceName)
```
fedepaol, line 828: Same for this test, this can use the generic policy

test/conformance/tests/sriov_operator.go
```diff
@@ -51,273 +51,13 @@ var _ = Describe("[sriov] operator", func() {
...
-							PfNames: []string{intf.Name + "#0-2"},
-						},
-						DeviceType: "vfio-pci",
-					},
-				}
-
-				err = clients.Create(context.Background(), secondConfig)
-				Expect(err).To(HaveOccurred())
-			})*/
+	Describe("Shared SriovNetworkNodePolicy", func() {
```
fedepaol, line 54: Can this be "Generic SriovNetworkNodePolicy" ?

test/conformance/tests/sriov_operator.go
```diff
@@ -681,292 +423,566 @@ var _ = Describe("[sriov] operator", func() {
...
-						NumVfs:       5,
-						ResourceName: "apivolresource",
-						Priority:     99,
-						NicSelector: sriovv1.SriovNetworkNicSelector{
-							PfNames: []string{intf.Name},
-						},
-						DeviceType: "netdevice",
-					},
-				}
+	Describe("Requiring new SriovNetworkNodePolicy", func() {
```
fedepaol, line 428: Can this be "Custom SriovNetworkNodePolicy" ?

test/util/namespaces/namespaces.go
mmirecki: Done

test/conformance/tests/sriov_operator.go
mmirecki: Done

test/conformance/tests/sriov_operator.go
mmirecki: Done

test/conformance/tests/sriov_operator.go
mmirecki: Done

test/conformance/tests/sriov_operator.go
mmirecki: Done

test/conformance/tests/sriov_operator.go
mmirecki: Done

test/conformance/tests/sriov_operator.go
```diff
@@ -681,292 +424,559 @@ var _ = Describe("[sriov] operator", func() {
...
+
+		BeforeEach(func() {
+			err := namespaces.Clean(operatorNamespace, namespaces.Test, clients)
+			Expect(err).ToNot(HaveOccurred())
+			waitForSRIOVStable()
+		})
+
+		Describe("Configuration", func() {
+
+			Context("SR-IOV network config daemon can be set by nodeselector", func() {
```
fedepaol, line 485: This does not need a particular node policy, can be moved in the other section

test/util/execute/discovery.go
```diff
@@ -0,0 +1,12 @@
+package execute
```
fedepaol, line 1: Can this be moved in the `discovery` package so it's consistent with the other test suites?

mmirecki: removed the file, it should only come in a following PR

fedepaol: Just curious, why can't you add it as part of this?

test/conformance/tests/sriov_operator.go
mmirecki: let's do this in another pr

test/util/execute/discovery.go
mmirecki: This code would be unused by this patch.
It's better to add it in the PR that will actually use it.

PR #233, opened 2020-06-03
Fix: Change mellanox total VFs to 0 when mellanox PF has 0 VFs

Mmduh-483:


pliurh:
Does it matter if we don't reset the NUM_OF_VFS in the firmware?

Mmduh-483:
> Does it matter if we don't reset the NUM_OF_VFS in the firmware?

For mellanox the Nics it is the intended behavior, if the numVFs is 0 then reset the total VFs in the firmware

pliurh:
> > Does it matter if we don't reset the NUM_OF_VFS in the firmware?
> 
> For mellanox Nics the intended behavioris to reset the total VFs in the firmware if numVFs is 0

I mean if we choose not to reset the NUM_OF_VFS and disable SRIOV in the firmware, is there any drawback?

Mmduh-483:
> I mean if we choose not to reset the NUM_OF_VFS and disable SRIOV in the firmware, is there any drawback?

Not sure if there is draw back, @moshe010 AFAIK it is mainly for saving resources, right?

adrianchiris:
> I mean if we choose not to reset the NUM_OF_VFS and disable SRIOV in the firmware, is there any drawback?

So during system boot the NIC requests from BIOS/UEFI to reserve system memory addresses. When SR-IOV is enabled it needs to request more.

It is roughly an addition of <max num of VFs>*<VF BAR size>

zshi-redhat:
nit: Should we only configure the firmware but not reboot the node? Just think from user perspective, is it expected that node be rebooted when deleting a policy?
Another example is, iommu kernel arguments are applied when vfio-pci mode is used, but there is no cleanup/reboot when policy is removed.

adrianchiris:
IMO this can be deferred to the next reboot.

Mmduh-483:
Changed the behavior to change the firmware without reboot, it can be applied in the next reboot

pkg/plugins/mellanox/mellanox_plugin.go
```diff
@@ -138,11 +138,31 @@ func (p *MellanoxPlugin) OnNodeStateChange(old, new *sriovnetworkv1.SriovNetwork
...
+		processedNics[pciPrefix] = true
+		pciAddress := pciPrefix + "0"
+
+		_, fwNext, err := getMlnxNicFwData(pciAddress)
+		if err != nil {
+			return false, false, err
+		}
+
+		if fwNext.totalVfs > 0 || fwNext.enableSriov {
+			attributesToChange[pciAddress] = mlnxNic{totalVfs: 0}
```
zshi-redhat, line 162: @Mmduh-483 please add a level 2 log here: `glog.V(2).Infof("Changing TotalVfs %d to 0, don't require rebooting", fwNext.totalVfs)`

pkg/plugins/mellanox/mellanox_plugin.go
```diff
@@ -338,6 +360,12 @@ func handleTotalVfs(fwCurrent, attrs *mlnxNic, ifaceSpec sriovnetworkv1.Interfac
 		needReboot = true
 	}
 
+	// Remove policy then re-apply it
+	if !needReboot && fwNext.totalVfs != totalVfs {
+		attrs.totalVfs = totalVfs
```
zshi-redhat, line 365: @Mmduh-483 please add a level 2 log here.

Mmduh-483:
@zshi-redhat Can you please re-review the PR

pkg/plugins/mellanox/mellanox_plugin.go
```diff
@@ -115,29 +115,47 @@ func (p *MellanoxPlugin) OnNodeStateChange(old, new *sriovnetworkv1.SriovNetwork
...
+		// Add the nic to processed Nics to not repeat the process for dual nic ports
+		processedNics[pciPrefix] = true
+		pciAddress := pciPrefix + "0"
+
+		_, fwNext, err := getMlnxNicFwData(pciAddress)
+		if err != nil {
+			return false, false, err
+		}
+
+		if fwNext.totalVfs > 0 || fwNext.enableSriov {
```
zshi-redhat, line 153: @Mmduh-483  nit: if fwNext.enableSriov is disabled, will it still be possible that fwNext.totalVfs is larger than 0?
My concern is when `fwNext.enableSriov` is used, it may try to configure devices that are not supported (for example, old CX-3 device which has enableSriov configured to true by default). and we don't guarantee this will work with mstconfig or any openshift platform.

Mmduh-483: Updated it to skip unsupported devices

zshi-redhat: @Mmduh-483 I just tried this combined settings `SRIOV_EN=false` && `NUM_OF_VFS=4` and found that they can be configured successfully to the firmware.  Is this expected? my understand was if SR-IOV_ENV is disabled, it's not possible to configure num_vfs.

For the fix, I was thinking we can simply remove the `fwNext.enableSriov` because if `fwNext.totalVfs > 0` is true, it means operator has successfully configured the device (which should be a supported device).

Mmduh-483: @zshi-redhat yes it is possible  to set `SRIOV_EN` to false and set the `NUM_OF_VFS` (total vfs), when set `SRIOV_EN` to false, the files `/sys/class/net/$netdev/device/sriov_numvfs` and `sriov_totalvfs` disappeared, so you can't create vfs.

adrianchiris: As a note:
`SRIOV_EN` configures the NIC to whether or not expose SR-IOV capability to the host
`NUM_OF_VFS` set the maximal number of VFS the NIC supports

The parameters dont depend on each other. which is a good thing or else you would have needed to perform two operations and possibly an additional reboot if they were.

zshi-redhat: @Mmduh-483  @adrianchiris  thanks for the note and explaination!

adrianchiris: Sure :)

PR #236, opened 2020-06-04
Link type selector

Mmduh-483:


zshi-redhat:
@Mmduh-483 as discussed, I think we can remove the linkType selector from operator API.

PR #237, opened 2020-06-04
Unbind/Bind VFs after assigning admin mac to set node_guid

Mmduh-483:
rdma_cm uses node_guid of VF for RoCE, but when VFs are created the node_guid is zero, assigning the administrator mac then unbind/bind the VF's driver can set the node_guid


zshi-redhat:
@Mmduh-483 would you please add a annotation indicating why we are unbind/bind drivers here and when this shall be removed?

Irrelevant to this PR,  we will think about how to have a vendor specific initialization layer for doing such vendor specific configuration if not in current vendor plugins.

Mmduh-483:
@zshi-redhat updated the commit message and the description of the PR

PR #239, opened 2020-06-08
bond-cni test suite

marzianor9:


fedepaol:
/close
Bond cni is not getting shipped with sriov-operator for the time being

PR #241, opened 2020-06-10
validate isRdma config in operator webhook

zshi-redhat:


pkg/webhook/validate.go
```diff
@@ -106,6 +106,15 @@ func staticValidateSriovNetworkNodePolicy(cr *sriovnetworkv1.SriovNetworkNodePol
 			}
 		}
 	}
+
+	// TODO: Add automatic correction in Operator mutation webhook
```
pliurh, line 110: Normally, an admission controller doesn't do the automatic correction. I prefer we only set the default value if a flag is not set.

zshi-redhat: removed.

This is related to one idea that we do intelligent analysis, we can discuss it later.

pliurh:
Could you also add a test to `pkg/webhook/validate_test.go`?

zshi-redhat:
> Could you also add a test to `pkg/webhook/validate_test.go`?

added

pliurh:
v

PR #242, opened 2020-06-10
run rebootNode in frontstage

zshi-redhat:
This fix a possible race condition that:
node be uncordoned while node is still in the process
of rebooting in rebootNode go routine.


zshi-redhat:
> With #245, we may not need this patch any more.

We can close this one if it's prove to be not necessary.
but I believe we should call the reboot Node the same way in both add and update handlers.

pliurh:
Close this PR, as the target code has been refactored.

PR #243, opened 2020-06-10
Generate random node and port guid for IB VFs for un-initialized VFs

Mmduh-483:
By default Mellanox InfiniBand links are initinialized with a zeroed
node and port GUID To perform RDMA traffic with RDMA CM needs to
initialize node and port GUID

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -4,6 +4,9 @@ import (
 	// "bytes"
 	"fmt"
 	"io/ioutil"
+	"math/rand"
+	"net"
+
 	// "net"
 	"os"
 	"os/exec"
@@ -185,7 +188,11 @@ func configSriovDevice(iface *sriovnetworkv1.Interface, ifaceStatus *sriovnetwor
 			glog.Errorf("configSriovDevice(): fail to set NumVfs for device %s", iface.PciAddress)
 			return err
 		}
-		if err = setVfsAdminMac(ifaceStatus); err != nil {
+		if strings.EqualFold(iface.LinkType, "IB") {
+			if err = setVfsGuid(iface); err != nil {
+				return err
+			}
+		} else if err = setVfsAdminMac(ifaceStatus); err != nil {
 			return err
 		}
 	}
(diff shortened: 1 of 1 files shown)


pkg/utils/utils.go
```diff
@@ -431,3 +438,51 @@ func getLinkType(ifaceStatus sriovnetworkv1.InterfaceExt) string {
...
+			return err
+		}
+		guid := generateRandomGuid()
+		if err := netlink.LinkSetVfNodeGUID(pfLink, vfID, guid); err != nil {
+			return err
+		}
+		if err := netlink.LinkSetVfPortGUID(pfLink, vfID, guid); err != nil {
+			return err
+		}
+		if err = Unbind(addr); err != nil {
```
zshi-redhat, line 466: please correct me if I'm wrong:
1) For RoCE VF device to get node or port guid, it needs to first set VF admin MAC, then unbind/bind driver
2) For IB or IPoIB VF device to get node or port guid, it needs to explicitly set the VF GUID , then unbind/bind driver

If the issue of `RDMA_CM cannot work without node or port guid` is fixed, then we can remove unbind/bind.

moshe010: It true, but for the IB we alway have to set port_guid  and unbind/bind the driver.
This is need for Infiniband to work not releated to rdma_cm
without ib-kubernetes we will do it in the operator. 
in the future if we decided that  ib-kubernetes is a must the we can remove it from here and let   ib-kubernete and ib-sriov-cni to set it.

zshi-redhat: For IB SR-IOV, can we set a random default guid in ib-sriov-cni if not provided by user or ib-kubernetes?

moshe010: we can do it, but I thought to make it similar to RoCE so it will be one time change.

zshi-redhat: ok, I think the initialization of node/port guid is needed for IB device where ib-sriov-cni might not be used. so it will not resolve the issue when adding in ib-sriov-cni for IB.

pkg/utils/utils.go
```diff
@@ -431,3 +438,51 @@ func getLinkType(ifaceStatus sriovnetworkv1.InterfaceExt) string {
...
+		}
+		if err = BindDefaultDriver(addr); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func generateRandomGuid() net.HardwareAddr {
```
zshi-redhat, line 477: This generates random guid for VFs everytime when node gets rebooted or another IB/Ether device gets configured(which result in node rebooting). Will this cause any issue in the infiniband network?

zshi-redhat:
Does the mtu setting in generic plugin apply to IB device?

Mmduh-483:
> Does the mtu setting in generic plugin apply to IB device?

Not all values for mtu can work with the IB devices, if the mtu value is not supported it will fail `write /sys/bus/pci/devices/0000:02:00.0/net/ib0/mtu: invalid argument`

Mmduh-483:
> Shall we also add a field in SriovNetworkNodeState for GUID? just like MAC address for each `VirtualFunction` instance.

It would be good, I think we can add it in another PR, @moshe010

pkg/utils/utils.go
Mmduh-483: I am not sure about, @moshe010 what do you think?

adrianchiris: I don't expect to be issues with that.
the PF reports its VGUIDs the SM should handle it (update/remove entries etc...).

adrianchiris:
> Does the mtu setting in generic plugin apply to IB device?
yes, for ipoib in datagram mode (the default mode) it is capped at 4k and it needs to be <= to the value that was set in the SM for the ipoib multicast group SM config.

adrianchiris:
@zshi-redhat  Are there any more open items ?

zshi-redhat:
> @zshi-redhat Are there any more open items ?

nope

/lgtm

PR #245, opened 2020-06-10
Use upstream drain lib

pliurh:
Additionally, this patch also moves the `uncordon` action before the `syncStatus` is set to `succeded`. So that node will only become schedulable after the sync is finished.


hack/build-plugins.sh
```diff
@@ -35,5 +35,5 @@ mkdir -p ${BIN_PATH}
 CGO_ENABLED=1
 
 echo "Building ${REPO}/pkg/plugins (${VERSION_OVERRIDE})"
-CGO_ENABLED=${CGO_ENABLED} GOOS=${GOOS} GOARCH=${GOARCH} go build -buildmode=plugin -mod=vendor -ldflags "${GLDFLAGS} -s -w" ${GOFLAGS} -o ${BIN_PATH}/plugins/$1_plugin.so ${REPO}/pkg/plugins/$1
-#CGO_ENABLED=${CGO_ENABLED} GOOS=${GOOS} GOARCH=${GOARCH} go build -buildmode=plugin -mod=vendor -gcflags='all=-N -l' -ldflags "${GLDFLAGS} -s -w" ${GOFLAGS} -o ${BIN_PATH}/plugins/$1_plugin.so ${REPO}/pkg/plugins/$1
+CGO_ENABLED=${CGO_ENABLED} GOOS=${GOOS} GOARCH=${GOARCH} go build -buildmode=plugin -ldflags "${GLDFLAGS} -s -w" ${GOFLAGS} -o ${BIN_PATH}/plugins/$1_plugin.so ${REPO}/pkg/plugins/$1
```
SchSeba, line 38: why you remove the `-mod=vendor` here?

PR #246, opened 2020-06-11
Bug 1839754: Remove required feilds from SriovNetworkNodeState CRD

pliurh:
OLM doesn't allow CRD to introduce new required fields during an upgrade.

### DIFF
--- a/manifests/4.6/sriov-network-operator-sriovnetworknodestate.crd.yaml
+++ b/manifests/4.6/sriov-network-operator-sriovnetworknodestate.crd.yaml
@@ -57,15 +57,10 @@ spec:
                           type: string
                         vfRange:
                           type: string
-                      required:
-                      - policyName
-                      - resourceName
                       type: object
                     type: array
                 required:
-                - name
                 - pciAddress
-                - vfGroups
                 type: object
               type: array
           type: object
--- a/deploy/crds/sriovnetwork.openshift.io_sriovnetworknodestates_crd.yaml
+++ b/deploy/crds/sriovnetwork.openshift.io_sriovnetworknodestates_crd.yaml
@@ -57,15 +57,10 @@ spec:
                           type: string
                         vfRange:
                           type: string
-                      required:
-                      - policyName
-                      - resourceName
                       type: object
                     type: array
                 required:
-                - name
                 - pciAddress
-                - vfGroups
                 type: object
               type: array
           type: object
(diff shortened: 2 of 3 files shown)


manifests/4.6/sriov-network-operator-sriovnetworknodestate.crd.yaml
```diff
@@ -57,15 +57,10 @@ spec:
                           type: string
                         vfRange:
                           type: string
-                      required:
-                      - policyName
-                      - resourceName
                       type: object
                     type: array
                 required:
-                - name
```
zshi-redhat, line 66: In which case, a interface doesn't have a name?

manifests/4.6/sriov-network-operator-sriovnetworknodestate.crd.yaml
```diff
@@ -57,15 +57,10 @@ spec:
                           type: string
                         vfRange:
                           type: string
-                      required:
-                      - policyName
```
zshi-redhat, line 61: Are we allowing empty `policyName` and `resourceName`? if resource name is empty, how will a pod request the resource?

zshi-redhat:
> OLM doesn't allow CRD to introduce new required fields during an upgrade.

Could you link to the OLM requirement?
If these required options are removed from CRD, will they be admitted in operator webhook?

manifests/4.6/sriov-network-operator-sriovnetworknodestate.crd.yaml
pliurh: > > OLM doesn't allow CRD to introduce new required fields during an upgrade.
> 
> Could you link to the OLM requirement?
> If these required options are removed from CRD, will they be admitted in operator webhook?

https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md#on-compatibility

You can also find more detail in the BZ comments.

manifests/4.6/sriov-network-operator-sriovnetworknodestate.crd.yaml
pliurh: SriovNetworkNodeState CRs are managed by the operator itself. So we shall not worry about the illegal configuration too much.

pliurh:
> > OLM doesn't allow CRD to introduce new required fields during an upgrade.
> 
> Could you link to the OLM requirement?
> If these required options are removed from CRD, will they be admitted in operator webhook?

https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md#on-compatibility

You can also find more detail in the BZ comments.

manifests/4.6/sriov-network-operator-sriovnetworknodestate.crd.yaml
pliurh: SriovNetworkNodeState CRs are managed by the operator itself. So we shall not worry about the illegal configuration too much.

PR #250, opened 2020-06-16
Add an optional label selector to limit the nodes impacted by the tests.

fedepaol:
By setting the optional NODES_SELECTOR env variable, we limit the nodes impacted by the tests to those matching that label selector.

### DIFF
--- a/test/conformance/tests/sriov_operator.go
+++ b/test/conformance/tests/sriov_operator.go
@@ -20,6 +20,7 @@ import (
 	"github.com/openshift/sriov-network-operator/test/util/execute"
 	"github.com/openshift/sriov-network-operator/test/util/namespaces"
 	"github.com/openshift/sriov-network-operator/test/util/network"
+	"github.com/openshift/sriov-network-operator/test/util/nodes"
 	"github.com/openshift/sriov-network-operator/test/util/pod"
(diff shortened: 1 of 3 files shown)


SchSeba:
/hold 

This needs to be validated

SchSeba:
> looks good but I think we also need to add this change in the vfs_assignement.go file

you add the selector under the `DiscoverSriov` so it should be fine :)

test/conformance/tests/sriov_operator.go
```diff
@@ -83,8 +84,13 @@ var _ = Describe("[sriov] operator", func() {
 				allNodes, err := clients.Nodes().List(metav1.ListOptions{
 					LabelSelector: "node-role.kubernetes.io/worker",
 				})
-				Expect(len(allNodes.Items)).To(BeNumerically(">", 0), "There must be at least one worker")
-				candidate := allNodes.Items[0]
+				Expect(err).ToNot(HaveOccurred())
+
+				nn, err := nodes.MatchingOptionalSelector(clients, allNodes.Items)
```
mmirecki, line 89: nn -> selectedNodes

fedepaol: done

PR #256, opened 2020-06-18
Export a clean.All function that removes all the leftovers from the conformance suite

fedepaol:
This is useful when using the suite as a dependency, to restore the cluster to the original state.

### DIFF
--- a/test/util/clean/clean.go
+++ b/test/util/clean/clean.go
@@ -1,0 +1,34 @@
+package clean
+
+import (
+	"fmt"
+	"os"
+	"time"
+
+	"github.com/openshift/sriov-network-operator/test/util/client"
+	"github.com/openshift/sriov-network-operator/test/util/namespaces"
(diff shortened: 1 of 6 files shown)


test/util/clean/clean.go
```diff
@@ -0,0 +1,25 @@
...
+
+	"github.com/openshift/sriov-network-operator/test/util/client"
+	"github.com/openshift/sriov-network-operator/test/util/namespaces"
+)
+
+// All cleans all the dangling resources created by conformance tests.
+// This includes pods, networks, policies and namespaces.
+func All() error {
+	clients := client.New("")
+	err := namespaces.Clean("openshift-sriov-network-operator", namespaces.Test, clients)
```
SchSeba, line 15: This should be a variable

SchSeba:
/lgtm

/hold

please remove the hold after validation

PR #260, opened 2020-06-19
Improve multi-node draining collision detection mechanism

pliurh:


zshi-redhat:
For dn.drainable flag, should we set defaut value to false ?
NodeUpdate and NodeStateChange are handlers for different informer, if for various reason NodeUpdate is not triggered, but NodeStateChange is called, will all config daemon start draining process (given the default value for dn.drainable is true)?

pkg/daemon/daemon.go
```diff
@@ -621,8 +615,8 @@ func (dn *Daemon) annotateNode(node, value string) error {
 func (dn *Daemon) drainNode(name string) {
 	glog.Info("drainNode(): Update prepared")
 	var err error
-
-	err = wait.PollImmediateUntil(3*time.Second, func() (bool, error) {
+	// wait a random time to avoid all the nodes drain at the same time
+	err = wait.PollUntil(time.Duration(rand.Intn(15)+1)*time.Second, func() (bool, error) {
```
zshi-redhat, line 619: If an error occurs when waiting for the pool, shall it return the err instead of continuing drain process below (annotate node etc)?

pliurh:
> For dn.drainable flag, should we set defaut value to false ?
> NodeUpdate and NodeStateChange are handlers for different informer, if for various reason NodeUpdate is not triggered, but NodeStateChange is called, will all config daemon start draining process (given the default value for dn.drainable is true)?

I would say it's a very corner case that you mentioned, as all the client-go objects share the same TCP connection. Also, it's nearly impossible that it happens on every node at the same time. Setting the default to `false` would not be helpful in such a case either.

pkg/daemon/daemon.go
pliurh: This poll is just waiting for the `drainalbe` to become `true`, so I don't think any error would happen here. The problem is that it will wait forever and block the configure daemon from handling any new requests.
However, in a large cluster, the waiting time could be very long. So I am not sure whether we should put a timeout here.

pliurh:
@zshi-redhat PTAL

PR #261, opened 2020-06-19
cleanup DaemonSet affinity when deleting plugin pods

zshi-redhat:
node affinity still exists when all policies are deleted.
try removing daemon affinity when deleting plugin pods.


pliurh:
/lgtm

BTW, without this patch, will there be any issue.

zshi-redhat:
> /lgtm
> 
> BTW, without this patch, will there be any issue.

Nope, there won't be issue as policy controller will explicitly set node selector to none OS for plugin daemons which prevent them from being deployed on any node.

PR #262, opened 2020-06-22
Not exit config daemon when sync failed

pliurh:
Use work queue to store update request, retry when any sync error happen

### DIFF
--- a/hack/run-test.sh
+++ b/hack/run-test.sh
@@ -11,4 +11,4 @@ export KUBECONFIG=${KUBECONFIG:-/root/dev-scripts/ocp/sriov/auth/kubeconfig}
 
 cd $DIR
 # GO111MODULE=on go test ./test/operator/...  -root=$OPERATOR_ROOT -kubeconfig=$KUBECONFIG -globalMan $OPERATOR_ROOT/deploy/crds/sriovnetwork.openshift.io_sriovnetworks_crd.yaml -namespacedMan $OPERATOR_ROOT/deploy/operator-init.yaml -v -singleNamespace true
-ginkgo -v --progress ./test/$1 -- -root=$DIR -kubeconfig=$KUBECONFIG -globalMan $DIR/hack/dummy.yaml -namespacedMan $DIR/hack/dummy.yaml -singleNamespace true
+ginkgo -v --progress ./test/$1 -- -root=$DIR -kubeconfig=$KUBECONFIG -globalMan $DIR/hack/dummy.yaml -namespacedMan $DIR/hack/dummy.yaml
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -349,30 +349,16 @@ func (dn *Daemon) operatorConfigChangeHandler(old, new interface{}) {
 }
 
 func (dn *Daemon) nodeStateSyncHandler(generation int64) error {
-	var err, lastErr error
+	var err error
 	glog.V(0).Infof("nodeStateSyncHandler(): new generation is %d", generation)
 	// Get the latest NodeState
 	var latestState *sriovnetworkv1.SriovNetworkNodeState
-	err = wait.PollImmediate(10*time.Second, 1*time.Minute, func() (bool, error) {
-		latestState, lastErr = dn.client.SriovnetworkV1().SriovNetworkNodeStates(namespace).Get(context.Background(), dn.name, metav1.GetOptions{})
-		if lastErr == nil {
-			return true, nil
-		}
-		glog.Warningf("nodeStateSyncHandler(): Failed to fetch node state %s (%v); retrying...", dn.name, lastErr)
-		return false, nil
-	})
+	latestState, err = dn.client.SriovnetworkV1().SriovNetworkNodeStates(namespace).Get(context.Background(), dn.name, metav1.GetOptions{})
 	if err != nil {
-		if err == wait.ErrWaitTimeout {
-			return fmt.Errorf("nodeStateSyncHandler(): Timed out trying to fetch the latest node state: %v", lastErr)
-		}
-		return fmt.Errorf("%v: %v", err, lastErr)
+		glog.Warningf("nodeStateSyncHandler(): Failed to fetch node state %s: %v", dn.name, err)
+		return err
 	}
 	latest := latestState.GetGeneration()
-	if latest != generation {
-		glog.Infof("nodeStateSyncHandler(): the latest generation is %d, skip", latest)
-		return nil
-	}
-
 	if dn.nodeState.GetGeneration() == latest {
 		glog.V(0).Infof("nodeStateSyncHandler(): Interface not changed")
 		return nil
@@ -473,18 +459,11 @@ func (dn *Daemon) nodeStateSyncHandler(generation int64) error {
 }
 
 func (dn *Daemon) completeDrain() error {
-	var lastErr error
-	err := wait.PollImmediate(10*time.Second, 1*time.Minute, func() (bool, error) {
-		if lastErr = drain.RunCordonOrUncordon(dn.drainer, dn.node, false); lastErr != nil {
-			return false, nil
-		}
-		return true, nil
-	})
-	if err != nil {
-		glog.Errorf("drainNode(): fail to uncordon node: %v: %v", err, lastErr)
+	if err := drain.RunCordonOrUncordon(dn.drainer, dn.node, false); err != nil {
 		return err
 	}
-	if err = dn.annotateNode(dn.name, annoIdle); err != nil {
+
+	if err := dn.annotateNode(dn.name, annoIdle); err != nil {
 		glog.Errorf("drainNode(): failed to annotate node: %v", err)
 		return err
 	}
@@ -497,54 +476,33 @@ func (dn *Daemon) restartDevicePluginPod() error {
 	glog.V(2).Infof("restartDevicePluginPod(): try to restart device plugin pod")
 
 	var podToDelete string
-	var foundPodToDelete bool
-	if err := wait.PollImmediateUntil(3*time.Second, func() (bool, error) {
-		pods, err := dn.kubeClient.CoreV1().Pods(namespace).List(context.Background(), metav1.ListOptions{
-			LabelSelector: "app=sriov-device-plugin",
-			FieldSelector: "spec.nodeName=" + dn.name,
-		})
-
+	pods, err := dn.kubeClient.CoreV1().Pods(namespace).List(context.Background(), metav1.ListOptions{
+		LabelSelector: "app=sriov-device-plugin",
+		FieldSelector: "spec.nodeName=" + dn.name,
+	})
+	if err != nil {
 		if errors.IsNotFound(err) {
 			glog.Info("restartDevicePluginPod(): device plugin pod exited")
-			return true, nil
+			return nil
 		}
-
-		if err != nil {
-			glog.Warningf("restartDevicePluginPod(): Failed to list device plugin pod: %s, retrying", err)
-			return false, nil
-		}
-
-		if len(pods.Items) == 0 {
-			glog.Info("restartDevicePluginPod(): device plugin pod exited")
-			return true, nil
-		}
-		podToDelete = pods.Items[0].Name
-		foundPodToDelete = true
-		return true, nil
-	}, dn.stopCh); err != nil {
-		glog.Errorf("restartDevicePluginPod(): failed to wait for finding pod to delete: %v", err)
+		glog.Warningf("restartDevicePluginPod(): Failed to list device plugin pod: %s, retrying", err)
 		return err
 	}
 
-	if !foundPodToDelete {
-		glog.Info("restartDevicePluginPod(): device plugin pod was not there")
+	if len(pods.Items) == 0 {
+		glog.Info("restartDevicePluginPod(): device plugin pod exited")
 		return nil
 	}
+	podToDelete = pods.Items[0].Name
 
-	if err := wait.PollImmediateUntil(3*time.Second, func() (bool, error) {
-		glog.V(2).Infof("restartDevicePluginPod(): Found device plugin pod %s, deleting it", podToDelete)
-		err := dn.kubeClient.CoreV1().Pods(namespace).Delete(context.Background(), podToDelete, metav1.DeleteOptions{})
-		if errors.IsNotFound(err) {
-			glog.Info("restartDevicePluginPod(): pod to delete not found")
-			return true, nil
-		}
-		if err != nil {
-			glog.Errorf("restartDevicePluginPod(): Failed to delete device plugin pod: %s, retrying", err)
-			return false, nil
-		}
-		return true, nil
-	}, dn.stopCh); err != nil {
-		glog.Errorf("restartDevicePluginPod(): failed to wait for pod deletion: %v", err)
+	glog.V(2).Infof("restartDevicePluginPod(): Found device plugin pod %s, deleting it", podToDelete)
+	err = dn.kubeClient.CoreV1().Pods(namespace).Delete(context.Background(), podToDelete, metav1.DeleteOptions{})
+	if errors.IsNotFound(err) {
+		glog.Info("restartDevicePluginPod(): pod to delete not found")
+		return nil
+	}
+	if err != nil {
+		glog.Errorf("restartDevicePluginPod(): Failed to delete device plugin pod: %s, retrying", err)
 		return err
 	}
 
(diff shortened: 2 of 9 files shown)


pkg/daemon/daemon.go
```diff
@@ -425,19 +442,19 @@ func (dn *Daemon) nodeStateChangeHandler(old, new interface{}) {
...
 		syncStatus:    "Succeeded",
 		lastSyncError: "",
 	}
+	return nil
 }
 
 func (dn *Daemon) completeDrain() {
 	if err := drain.RunCordonOrUncordon(dn.drainer, dn.node, false); err != nil {
-		glog.Errorf("nodeStateChangeHandler(): fail to uncordon node: %v", err)
+		glog.Errorf("nodeStateSyncHandler(): fail to uncordon node: %v", err)
```
zshi-redhat, line 457: Remove the `dn.exitCh` when `annotateNode` or `RunCordonOrUncordon` fail and add a error check after `dn.completeDrain`.

pkg/daemon/daemon.go
```diff
@@ -251,145 +339,76 @@ func (dn *Daemon) operatorConfigChangeHandler(old, new interface{}) {
...
 		}
-		dn.exitCh <- err
-		return
+		return err
 	}
-	latest := latestState.GetObjectMeta().GetGeneration()
-	if latest != newState.GetObjectMeta().GetGeneration() {
-		glog.Errorf("nodeStateChangeHandler(): the lastet generation is %d, skip", latest)
-		return
+	latest := latestState.GetGeneration()
```
fedepaol, line 370: Not sure, but could this logic be moved directly into the handler, before putting the event in the queue?

fedepaol:
Nice! :+1:

pkg/daemon/daemon.go
pliurh: Yes, it can be done. I did consider it. But a request may be put back into the queue when 1) the node state changes 2) an error happens in the handler. So if we choose to move the logic, then we need to put it to both places, which I tend to not.

fedepaol: So, this made me think to another issue.
If you put the event back on the queue for some error, you risk to have an older event to be consumed AFTER a more recent one. This is because in this patch we are putting the event itself and not just the event key (normally name + ns). 

The established pattern is to put the key of the event in the queue, and fetch the latest element from the informer cache when the key is read from the queue. This avoids the issue mentioned above (and probably many others), because you always consume the freshest.

In this case it does not come for free, because there are a few places where we use the old version (but not as many as I thought):
- The version check, that can be moved into the handler
- The `needDrain` function in generic plugin, that could check directly the value from sysfs
- The version of the DP configmap. This is probably the trickier one, which requires a bit more thought, because there's no easy way to understand if the running dp is using the current configmap or an old one. This also means that if we loose that change of dp configmap version event, we will never restart the dp.

I am afraid that relying on the old / new couple in the exit of the workqueue will leave us with new corner cases.

pkg/daemon/daemon.go
pliurh: Fixed.

pkg/daemon/daemon.go
pliurh: > The needDrain function in generic plugin, that could check directly the value from sysfs

Yes, comparing with sysfs is more reliable. But I think we can do it in a separate PR.

> relying on the old / new couple in the exit of the workqueue will leave us with new corner cases.

I agree. And I made some changes in the latest patch to reduce the dependency on the old object. So I think from the plugins' perspective, we don't need it anymore. I think I have to solution to solve the DP config problem. Let me refactor this patch a bit.

/hold

pliurh:
@zshi @fedepaol @SchSeba PTAL

pkg/daemon/daemon.go
```diff
@@ -398,54 +435,65 @@ func (dn *Daemon) nodeStateChangeHandler(old, new interface{}) {
...
+		if err := dn.completeDrain(); err != nil {
+			glog.Errorf("drainNode(): failed to complete draining: %v", err)
+			return err
+		}
 	} else if !ok {
 		if err := dn.annotateNode(dn.name, annoIdle); err != nil {
 			glog.Errorf("drainNode(): failed to annotate node: %v", err)
-			dn.exitCh <- err
-			return
+			return err
```
zshi-redhat, line 466: In which case, will this `else if` be executed?

pliurh: Before the node is annotated.

pkg/daemon/daemon.go
```diff
@@ -251,145 +353,80 @@ func (dn *Daemon) operatorConfigChangeHandler(old, new interface{}) {
...
-		return
+		return fmt.Errorf("%v: %v", err, lastErr)
 	}
-	latest := latestState.GetObjectMeta().GetGeneration()
-	if latest != newState.GetObjectMeta().GetGeneration() {
-		glog.Errorf("nodeStateChangeHandler(): the lastet generation is %d, skip", latest)
-		return
+	latest := latestState.GetGeneration()
+	if latest != generation {
+		glog.Infof("nodeStateSyncHandler(): the lastet generation is %d, skip", latest)
```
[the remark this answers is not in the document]
pliurh: We already have `new generation is` at the beginning of the function, that is the skipped generation.

pkg/daemon/daemon.go
```diff
@@ -398,54 +435,65 @@ func (dn *Daemon) nodeStateChangeHandler(old, new interface{}) {
...
 		}
 	}
+	glog.Info("nodeStateSyncHandler(): sync succeeded")
+	dn.nodeState = latestState.DeepCopy()
 	dn.refreshCh <- Message{
 		syncStatus:    "Succeeded",
 		lastSyncError: "",
 	}
+	// wait for writer to refresh the status
+	<-dn.syncCh
```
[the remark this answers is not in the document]
pliurh: This is for the scenario of https://bugzilla.redhat.com/show_bug.cgi?id=1815039

As in the `nodeStateSyncHandler`, we only check the generation of the CR. However, the generation can only reflect the change in the `spec` of CR. Without this sync, if users re-apply a policy right after deleting it, there is a change that the spec and status in the `latestState` are all in the desired state, as the deletion is not finished, the status is yet up-to-date. It will make the `needUpdate` function will return `false`. 

This sync blocks the change handler until the writer update the status. So that the latest state in invoking will not has such wrong information.

I don't think we need this sync when setting status to `InProgress`, as `nodeStateSyncHandler` doesn't exit at that time. But I think we may need it for the `Failed` case as well.

pkg/daemon/daemon.go
```diff
@@ -206,6 +256,60 @@ func (dn *Daemon) Run(stopCh <-chan struct{}, exitCh <-chan error) error {
 	}
 }
 
+func (dn *Daemon) runWorker() {
+	for dn.processNextWorkItem() {
+	}
+}
+
+func (dn *Daemon) processNextWorkItem() bool {
+	glog.Infof("worker queue size: %d", dn.workqueue.Len())
```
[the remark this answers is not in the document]
pliurh: Done

pkg/daemon/daemon.go
```diff
@@ -206,6 +256,60 @@ func (dn *Daemon) Run(stopCh <-chan struct{}, exitCh <-chan error) error {
...
+		// Finally, if no error occurs we Forget this item so it does not
+		// get queued again until another change happens.
+		dn.workqueue.Forget(obj)
+		glog.Infof("Successfully synced")
+		return nil
+	}(obj)
+
+	if err != nil {
+		utilruntime.HandleError(err)
+		return true
```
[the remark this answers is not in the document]
pliurh: Done

pkg/daemon/daemon.go
```diff
@@ -206,6 +256,60 @@ func (dn *Daemon) Run(stopCh <-chan struct{}, exitCh <-chan error) error {
...
 
+func (dn *Daemon) runWorker() {
+	for dn.processNextWorkItem() {
+	}
+}
+
+func (dn *Daemon) processNextWorkItem() bool {
+	glog.Infof("worker queue size: %d", dn.workqueue.Len())
+	obj, shutdown := dn.workqueue.Get()
+	glog.Infof("get item: %v", obj.(workItem).new.GetGeneration())
```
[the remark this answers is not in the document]
pliurh: Done

pkg/daemon/daemon.go
```diff
@@ -251,145 +355,80 @@ func (dn *Daemon) operatorConfigChangeHandler(old, new interface{}) {
...
 		}
-		glog.Warningf("nodeStateChangeHandler(): Failed to fetch node state %s (%v); retrying...", dn.name, lastErr)
+		glog.Warningf("nodeStateSyncHandler(): Failed to fetch node state %s (%v); retrying...", dn.name, lastErr)
 		return false, nil
 	})
 	if err != nil {
 		if err == wait.ErrWaitTimeout {
-			dn.exitCh <- goerr.Wrapf(lastErr, "nodeStateChangeHandler(): Timed out trying to fetch the latest node state %s", dn.name)
-			return
+			return goerr.Wrapf(lastErr, "nodeStateSyncHandler(): Timed out trying to fetch the latest node state %s", dn.name)
```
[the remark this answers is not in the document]
pliurh: Done

pkg/daemon/daemon.go
```diff
@@ -251,145 +355,80 @@ func (dn *Daemon) operatorConfigChangeHandler(old, new interface{}) {
...
 		}
-		dn.exitCh <- err
-		return
+		return err
```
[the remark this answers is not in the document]
pliurh: Done

pkg/daemon/daemon.go
```diff
@@ -398,54 +435,65 @@ func (dn *Daemon) nodeStateChangeHandler(old, new interface{}) {
...
-			return
+			glog.Errorf("nodeStateSyncHandler(): fail to restart device plugin pod: %v", err)
+			return err
 		}
 	}
 
 	if anno, ok := dn.node.Annotations[annoKey]; ok && anno == annoDraining {
-		dn.completeDrain()
+		if err := dn.completeDrain(); err != nil {
+			glog.Errorf("drainNode(): failed to complete draining: %v", err)
```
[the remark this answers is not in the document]
pliurh: Done

pkg/daemon/daemon.go
```diff
@@ -398,54 +435,65 @@ func (dn *Daemon) nodeStateChangeHandler(old, new interface{}) {
...
+func (dn *Daemon) completeDrain() error {
+	var lastErr error
+	err := wait.PollImmediate(10*time.Second, 1*time.Minute, func() (bool, error) {
+		if lastErr = drain.RunCordonOrUncordon(dn.drainer, dn.node, false); lastErr != nil {
+			return false, nil
+		}
+		return true, nil
+	})
+	if err != nil {
+		glog.Errorf("drainNode(): fail to uncordon node: %v: %v", err, lastErr)
```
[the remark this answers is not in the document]
pliurh: Done

pkg/daemon/daemon.go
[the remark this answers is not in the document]
zshi-redhat: ack

pkg/daemon/daemon.go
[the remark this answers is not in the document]
zshi-redhat: ack

pkg/daemon/daemon.go
```diff
@@ -245,145 +348,80 @@ func (dn *Daemon) operatorConfigChangeHandler(old, new interface{}) {
...
 	}
-	glog.V(2).Infof("nodeStateChangeHandler(): reqDrain %v, reqReboot %v", reqDrain, reqReboot)
+	glog.V(0).Infof("nodeStateSyncHandler(): reqDrain %v, reqReboot %v", reqDrain, reqReboot)
 
 	if reqDrain {
-		glog.Info("nodeStateChangeHandler(): drain node")
-		dn.drainNode(latestState.GetName())
+		glog.Info("nodeStateSyncHandler(): drain node")
+		if err := dn.drainNode(dn.name); err != nil {
+			return err
```
zshi-redhat, line 416: When dn.drainNode fails, shall it call completeDrain and other nodes may get chance to start draining?
This node, when drainNode fails, will be marked with syncError (draining in this node completes with error).
This may avoid draining process be blocked by one error node.

pkg/daemon/daemon.go
zshi-redhat: Every node will first be annotated as Draining, then be marked as Idle, how would the first time annotating happen?

pkg/daemon/daemon.go
```diff
@@ -206,6 +244,71 @@ func (dn *Daemon) Run(stopCh <-chan struct{}, exitCh <-chan error) error {
 	}
 }
 
+func (dn *Daemon) runWorker() {
+	for dn.processNextWorkItem() {
+	}
+}
+
+func (dn *Daemon) enqueueFoo(obj interface{}) {
```
zshi-redhat, line 252: Maybe use a descriptive name, such as enqueueNodeState. Foo is like an example.

pkg/daemon/daemon.go
```diff
@@ -245,145 +348,80 @@ func (dn *Daemon) operatorConfigChangeHandler(old, new interface{}) {
...
-		dn.exitCh <- err
-		return
+		return fmt.Errorf("%v: %v", err, lastErr)
 	}
-	latest := latestState.GetObjectMeta().GetGeneration()
-	if latest != newState.GetObjectMeta().GetGeneration() {
-		glog.Errorf("nodeStateChangeHandler(): the lastet generation is %d, skip", latest)
-		return
+	latest := latestState.GetGeneration()
+	if latest != generation {
```
fedepaol, line 371: This can be moved in the updateHandler, so you end up having only different versions in the queue.
Also, note that the final order you process them might be different from the original one, because if one error happens, the event is put back in the queue and processed after a fresher one, but I don't think this is an issue.

pkg/daemon/daemon.go
```diff
@@ -245,145 +348,80 @@ func (dn *Daemon) operatorConfigChangeHandler(old, new interface{}) {
...
 		}
-		glog.Warningf("nodeStateChangeHandler(): Failed to fetch node state %s (%v); retrying...", dn.name, lastErr)
+		glog.Warningf("nodeStateSyncHandler(): Failed to fetch node state %s (%v); retrying...", dn.name, lastErr)
 		return false, nil
 	})
 	if err != nil {
 		if err == wait.ErrWaitTimeout {
-			dn.exitCh <- goerr.Wrapf(lastErr, "nodeStateChangeHandler(): Timed out trying to fetch the latest node state %s", dn.name)
-			return
+			return fmt.Errorf("nodeStateSyncHandler(): Timed out trying to fetch the latest node state: %v", lastErr)
```
fedepaol, line 366: In theory, now we can just raise the error if the get fails, and the queue will handle the retry for us.

pkg/plugins/generic/generic_plugin.go
```diff
@@ -83,7 +83,7 @@ func (p *GenericPlugin) OnNodeStateChange(old, new *sriovnetworkv1.SriovNetworkN
 	err = nil
 	p.DesireState = new
 
-	needDrain = needDrainNode(new.Spec.Interfaces, old.Status.Interfaces)
+	needDrain = needDrainNode(new.Spec.Interfaces, new.Status.Interfaces)
```
fedepaol, line 86: This is going to work because the interfaces in the status part are updated asynchronously, right? And in any case, not in the version we are handling now.

hack/run-test.sh
```diff
@@ -11,4 +11,4 @@ export KUBECONFIG=${KUBECONFIG:-/root/dev-scripts/ocp/sriov/auth/kubeconfig}
 
 cd $DIR
 # GO111MODULE=on go test ./test/operator/...  -root=$OPERATOR_ROOT -kubeconfig=$KUBECONFIG -globalMan $OPERATOR_ROOT/deploy/crds/sriovnetwork.openshift.io_sriovnetworks_crd.yaml -namespacedMan $OPERATOR_ROOT/deploy/operator-init.yaml -v -singleNamespace true
-ginkgo -v --progress ./test/$1 -- -root=$DIR -kubeconfig=$KUBECONFIG -globalMan $DIR/hack/dummy.yaml -namespacedMan $DIR/hack/dummy.yaml -singleNamespace true
+ginkgo -v --progress ./test/$1 -- -root=$DIR -kubeconfig=$KUBECONFIG -globalMan $DIR/hack/dummy.yaml -namespacedMan $DIR/hack/dummy.yaml
```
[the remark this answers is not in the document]
pliurh: Done

pkg/daemon/daemon.go
pliurh: When the operator is deployed, the node will be annotated to Idle for the first time.

pkg/plugins/generic/generic_plugin.go
pliurh: The status is not always updated asynchronously, as in this patch I add a `syncCh` channel to block the daemon until writer updates the status. This is to make sure in each update handling, the config daemon can get the latest status which is modified by last round.

pkg/daemon/daemon.go
pliurh: hmmm, it makes sense. I need to do some tests.

pkg/daemon/daemon.go
pliurh: If we move it to the updateHandler, we shall do the same thing when putting a failed event back to the queue as well. After all, we don't really care about the key itself, we only need to keep the state synced with the latest. Therefore I prefer to leave this logic here.

pkg/daemon/daemon.go
pliurh: When draining fails, it most likely caused by a lack of spare resources. I don't think we shall continue to drain other nodes in such a case.

pkg/daemon/daemon.go
pliurh: done

pkg/daemon/daemon.go
pliurh: I removed some retries, it works. However, I cannot see the API timeout error in the latest 4.6 build anymore.

pkg/daemon/daemon.go
zshi-redhat: Lack of resources in one node doesn't mean other nodes are not drainable. Take a fat pod as an example, it may not be evicted to another node due to resource constraints, but thin pods on other nodes might be able to migrate to this node or any other nodes(except the fat pod node).

zshi-redhat: Before adding this patch, I think we agree on the need of a drain timeout, which allows the drain process on the node to proceed. 
After introducing this patch, if we don't complete the drain process upon failure in one node, it will stuck forever.

pkg/daemon/daemon.go
zshi-redhat: Ok, would it make sense to move the annotating action into completeDrain? this will make annotating associated with draining process, e.g annotating `draining` in `drainNode` and annotating `idle` in `completeDrain`.

pliurh: > Ok, would it make sense to move the annotating action into completeDrain? this will make annotating associated with draining process, e.g annotating `draining` in `drainNode` and annotating `idle` in `completeDrain`.

This is the current implementable already. The `annotateNode` here is just for setting the annotation to the initial value.

pkg/daemon/daemon.go
pliurh: There is a timeout in draining. If the draining fails, the event will be put back to the queue and retry. Yes, the node may be stuck in 'unschedulable' forever. But if we `uncordon` this node and give other nodes the opportunity to be drained, and the draining fails on other nodes. Then multiple nodes may be stuck in a drain-failed-uncordon cycle, which could affect more than one node.

zshi-redhat: Just a note: the draining may or may not fail on other nodes, the failure of drain in the affected node does not necessarily affect non-sriov components from running on the same node or other nodes. 

It's that for current implementation, we choose to block other nodes from draining in case any one of cluster nodes fails.

pkg/daemon/daemon.go
fedepaol: If you move to `updateHandler` you'll have the guarantee of having only different generations in the queue.
Also, this if `if latest != generation {` may be true if you are triggered by a variation, then you read the current status using the client and you retrieve the very same generation.

pkg/plugins/generic/generic_plugin.go
fedepaol: Ah, I missed that! Thanks!

pkg/daemon/daemon.go
fedepaol: It would make sense to understand if the drain failed because pods could not be evicted (in that case it does not make sense to retry) or because of some error that is worth retrying (such as connection error with the apiserver).
But this would probably be too complex and error prone. 
I see @zshi-redhat 's point, if there's a fat pod, having some nodes drained or not is a matter of the order we receive the states. How about the initial idea of `completeDrain` and returning an error? In that case, on the long run if a node is not able to drain won't affect the others.
Also, note that in `drainNode` we don't check if the wait for the node to be drainable or not succeded or failed. Not sure it's wanted or not.

fedepaol:
LGTM, leaving the last approval to @zshi-redhat 
I guess we could investigate the draining logic afterwards

zshi-redhat:
> LGTM, leaving the last approval to @zshi-redhat

thanks @fedepaol 
/lgtm

> I guess we could investigate the draining logic afterwards

agree

PR #263, opened 2020-06-24
partition: merge PF configs

zshi-redhat:
PF configs may affect how VF works, for example:
If VF mtu is set to a value larger than PF mtu, application
will not benefit from large mtu size configured for VF. This
happens when SR-IOV partition is configured with different mtu
sizes.

This commit also uses large numVfs in SR-IOV partition case.


pliurh:
I think we shall also consider the `priority`. This `mergePfConfigs` shall only merge policy with the same priority.

zshi-redhat:
> I think we shall also consider the `priority`. This `mergePfConfigs` shall only merge policy with the same priority.

right, added priority into consideration.

PR #264, opened 2020-06-24
WIP. Add matrix connectivity tests

kononovn:


test/util/connectivity/connectivity.go
```diff
@@ -0,0 +1,57 @@
+package connectivity
+
+import "fmt"
+
+// ConnectivityTest .
+type ConnectivityTest struct {
```
fedepaol, line 6: I would keep this and `Parameter` type safe and change only the rendering. It will make everything else easier.
So, a struct that contains the exact stuff you want (protocol, mtu, connectivity) and a way to render it in a format you like.

test/util/connectivity/connectivity.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+// RunTest starts connecitivity tests based on given parameters.
+func (c ConnectivityTest) RunTest() bool {
+	fmt.Println("Run Tests")
+	return true
+}
+
+// EnvValidation validates if it's possible to run specfic connectivity test on given environment.
+func (c ConnectivityTest) EnvValidation() bool {
```
fedepaol, line 41: I am afraid this is a bit romantic :-)
What is going to happen is that for each set of parameters, you need to find a node (or a resourceid, or something else) you want to use / run the pod on.

SchSeba:
@kononovn I think we can close the PR right?

SchSeba:
@kononovn ping can we close this PR?

PR #265, opened 2020-06-29
Set default values for non-required policy fields

zshi-redhat:


pliurh:
For MTU, if the link_type is IB, shall we also set 1500 as the default value?

zshi-redhat:
> For MTU, if the link_type is IB, shall we also set 1500 as the default value?

From RHEL doc, the common InfiniBand link-layer MTU is 2048.
We always set 1500 in Operator when resetting SR-IOV device for both eth/ib devices, the change in this PR is consistent with current Operator behavior.

For infiniband MTU, I think we can use 2048 as default one. I'll update the pr for webhook, and create a separate one for sriov resetting mtu change.

PR #267, opened 2020-06-29
Detect partition overlap regardless of rootDevices

zshi-redhat:
Previously partition overlap was rejected only when
policy.Spec.NicSelector.RootDevices is configured.


pliurh:
Could you also modify the tests in validate_test.go?

zshi-redhat:
> Could you also modify the tests in validate_test.go?

Done

PR #270, opened 2020-06-30
Fix hack/update-codegen.sh

zshi-redhat:
1. add --go-head-file for missing boilerplate error
2. restore generate-groups.sh file permission
3. update sriovnetwork/v1/zz_generated.deepcopy.go header


hack/update-codegen.sh
```diff
@@ -4,4 +4,6 @@ chmod +x vendor/k8s.io/code-generator/generate-groups.sh
 ./vendor/k8s.io/code-generator/generate-groups.sh all \
       github.com/openshift/sriov-network-operator/pkg/client \
       github.com/openshift/sriov-network-operator/pkg/apis \
-      sriovnetwork:v1
+      sriovnetwork:v1 \
+      --go-header-file hack/boilerplate.go.txt
+chmod -x vendor/k8s.io/code-generator/generate-groups.sh
```
SchSeba, line 9: we already have this line here https://github.com/openshift/sriov-network-operator/pull/270/files#diff-5345a009f642e55fc0d5b16d4f5d6f9bR3

zshi-redhat: @SchSeba I opened this link, but it didn't switch to a certain line. could you point me to the duplicate code?

zshi-redhat: For line 9, I remove the execution persmission from the generate-groups.sh file that was previously granted before executing the script. This avoid developer from wrongly submitting the file permission change.

hack/boilerplate.go.txt
```diff
@@ -0,0 +1,16 @@
+/*
+Copyright The Kubernetes Authors.
```
pliurh, line 2: Do we need this line?

zshi-redhat: This line was included in many pkg/client files already.
I think we can change it if necessary.

zshi-redhat:
@SchSeba @pliurh any further comments on this one?

hack/boilerplate.go.txt
pliurh: As the sriov operator is not part of the k8s project but part of openshift project, so I think maybe we should change it to something like `The Openshift Authors` instead.

zshi-redhat: so we will change all the affected files in sriov operator repo?

PR #275, opened 2020-07-01
Add discovery mode for sriov tests with shared policy

mmirecki:
This patch adds a check for discovery mode and a resource check for tests with a shared policy.
Other tests will be covered in a follow up patch.

### DIFF
--- a/test/conformance/tests/sriov_operator.go
+++ b/test/conformance/tests/sriov_operator.go
@@ -17,7 +17,9 @@ import (
 	netattdefv1 "github.com/openshift/sriov-network-operator/pkg/apis/k8s/v1"
 	sriovv1 "github.com/openshift/sriov-network-operator/pkg/apis/sriovnetwork/v1"
 	"github.com/openshift/sriov-network-operator/test/util/cluster"
+	"github.com/openshift/sriov-network-operator/test/util/discovery"
 	"github.com/openshift/sriov-network-operator/test/util/execute"
+
 	"github.com/openshift/sriov-network-operator/test/util/namespaces"
 	"github.com/openshift/sriov-network-operator/test/util/network"
 	"github.com/openshift/sriov-network-operator/test/util/nodes"
@@ -71,7 +73,6 @@ var _ = Describe("[sriov] operator", func() {
 	})
 
 	Describe("No SriovNetworkNodePolicy", func() {
-
 		Context("SR-IOV network config daemon can be set by nodeselector", func() {
 			// 26186
 			It("Should schedule the config daemon on selected nodes", func() {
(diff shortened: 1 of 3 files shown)


test/util/discovery/discovery.go
```diff
@@ -0,0 +1,72 @@
...
+	for _, node := range sriovNodes {
+		if node == name {
+			return true
+		}
+	}
+	return false
+}
+
+// DiscoveredResources discovers resources needed by the tests in discovery mode
+func DiscoveredResources(clients *client.ClientSet, sriovNodes []string, operatorNamespace string) (preferedNode string, preferedResourceName string, preferedResourceCount int, err error) {
```
fedepaol, line 47: Nit: "preferredNode"

test/util/discovery/discovery.go
```diff
@@ -0,0 +1,72 @@
...
+	err = clients.List(context.Background(), &policyList, runtimeclient.InNamespace(operatorNamespace))
+	if err != nil {
+		return "", "", 0, err
+	}
+	nodes, err := getSriovNodes(clients, sriovNodes)
+	if err != nil {
+		return "", "", 0, err
+	}
+
+	for _, policy := range policyList.Items {
```
fedepaol, line 58: Does it make sense to pass a policyMatcher function here? In this first step we care about "vanilla" policies only, not esotic ones. Like, regular driver, no MTU, etc.

test/util/discovery/discovery.go
```diff
@@ -0,0 +1,72 @@
...
+	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// Enabled indicates whether test discovery mode is enabled.
+func Enabled() bool {
+	discoveryMode, _ := strconv.ParseBool(os.Getenv("DISCOVERY_MODE"))
+	return discoveryMode
+}
+
+func getSriovNodes(clients *client.ClientSet, sriovNodes []string) ([]corev1.Node, error) {
```
fedepaol, line 21: How about just iterating sriovNodes and Get-ing each node?

test/util/discovery/discovery.go
```diff
@@ -0,0 +1,72 @@
...
+// DiscoveredResources discovers resources needed by the tests in discovery mode
+func DiscoveredResources(clients *client.ClientSet, sriovNodes []string, operatorNamespace string) (preferedNode string, preferedResourceName string, preferedResourceCount int, err error) {
+	policyList := sriovv1.SriovNetworkNodePolicyList{}
+	err = clients.List(context.Background(), &policyList, runtimeclient.InNamespace(operatorNamespace))
+	if err != nil {
+		return "", "", 0, err
+	}
+	nodes, err := getSriovNodes(clients, sriovNodes)
+	if err != nil {
+		return "", "", 0, err
```
fedepaol, line 55: Either use named returns or not. I prefer not using them, but it's just a matter of taste.
Here you can just say "return" because err is set and the others are 0 values.

test/util/discovery/discovery.go
```diff
@@ -0,0 +1,72 @@
...
+	var nodes []corev1.Node
+	for _, node := range allNodes.Items {
+		if containsNode(node.ObjectMeta.Name, sriovNodes) {
+			nodes = append(nodes, node)
+		}
+	}
+	return nodes, nil
+}
+
+func containsNode(name string, sriovNodes []string) bool {
```
fedepaol, line 37: In any case, can you put this and all other "util functions" below `DiscoverResources` ? As they are secondary and used by `DiscoverResources`, it's fine to find them after in reading order.

test/conformance/tests/sriov_operator.go
```diff
@@ -139,61 +140,28 @@ var _ = Describe("[sriov] operator", func() {
...
-			// if we don't find a Mellanox card we fall back to any sriov
-			// capability interface and skip the rate limit test.
-			intf, err := sriovInfos.FindOneMellanoxSriovDevice(node)
-			if err != nil {
-				intf, err = sriovInfos.FindOneSriovDevice(node)
-				Expect(err).ToNot(HaveOccurred())
-			}
+			if discovery.Enabled() {
+				node, resourceName, numVfs, err = discovery.DiscoveredResources(clients, sriovInfos.Nodes, operatorNamespace)
+				discoveryModeSkip = node == "" || resourceName == "" || numVfs < 5 || err != nil
```
fedepaol, line 150: Expect err to not be nil.
Also, makes sense to provide a reason for why we skipped

test/conformance/tests/sriov_operator.go
```diff
@@ -139,61 +140,28 @@ var _ = Describe("[sriov] operator", func() {
...
-					},
-					DeviceType: "netdevice",
-				},
+			} else {
+				node = sriovInfos.Nodes[0]
+				createSriovNetworkNodePolicy(node, sriovInfos, numVfs, resourceName)
 			}
-
-			err = clients.Create(context.Background(), config)
+			sriovDevice, err = sriovInfos.FindOneSriovDevice(node)
```
fedepaol, line 156: This might not be enough, you will probably care about finding one of the "best resource"

test/conformance/tests/sriov_operator.go
```diff
@@ -139,61 +140,28 @@ var _ = Describe("[sriov] operator", func() {
 		resourceName := "testresource"
 		var node string
 		var sriovDevice *sriovv1.InterfaceExt
+		var discoveryModeSkip bool
```
fedepaol, line 143: discoveryFailed?

test/conformance/tests/sriov_operator.go
```diff
@@ -1596,3 +1564,49 @@ func waitForSRIOVStable() {
 		return isClusterReady
 	}, waitingTime, 1*time.Second).Should(BeTrue())
 }
+
+func createSriovNetworkNodePolicy(node string, sriovInfos *cluster.EnabledNodes, numVfs int, resourceName string) {
```
fedepaol, line 1568: createVanillaNetworkPolicy ?
Also, I would move `waitForSriovStable` outside, it goes beyond the purpose of this function.

test/util/discovery/discovery.go
mmirecki: Done,

test/util/discovery/discovery.go
mmirecki: Done

test/util/discovery/discovery.go
mmirecki: Yes! But this comes in the follow up PR, which wil take care of the non-shared policies.

fedepaol: Yep, but for the sakeness of this part we must be sure to pick one compatible.

test/conformance/tests/sriov_operator.go
mmirecki: Done,
Later we should also make sure that it picks one that the policy can actually use (for example by considering the policy nic selector), but this is outside the scope of this PR

test/conformance/tests/sriov_operator.go
mmirecki: Done

test/conformance/tests/sriov_operator.go
mmirecki: Done

test/util/discovery/discovery.go
mmirecki: Done

PR #272, opened 2020-06-30
add clientsets for sriovnetwork and sriovnetworknodepolicy

zshi-redhat:
Fixes #271

PR #276, opened 2020-07-02
Webhook respect unapplied policy

zshi-redhat:
Webhook validation against policy may be inaccurate
when previous defined poilcies are not yet converted to
SriovNetworkNodeState.

### DIFF
--- a/pkg/webhook/validate.go
+++ b/pkg/webhook/validate.go
@@ -127,12 +127,24 @@ func dynamicValidateSriovNetworkNodePolicy(cr *sriovnetworkv1.SriovNetworkNodePo
 	if err != nil {
 		return false, err
 	}
+	npList, err := snclient.SriovnetworkV1().SriovNetworkNodePolicies(namespace).List(context.Background(), metav1.ListOptions{})
+	if err != nil {
+		return false, err
+	}
 	for _, node := range nodeList.Items {
 		if cr.Selected(&node) {
 			nodesSelected = true
 			for _, ns := range nsList.Items {
 				if ns.GetName() == node.GetName() {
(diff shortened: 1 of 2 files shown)


zshi-redhat:
This PR has a dependency on the clientsets change introduced in https://github.com/openshift/sriov-network-operator/pull/272

/retest

pkg/webhook/validate.go
```diff
@@ -166,22 +178,35 @@ func validatePolicyForNodeState(policy *sriovnetworkv1.SriovNetworkNodePolicy, s
...
-						}
-					}
+	for _, curPf := range current.Spec.NicSelector.PfNames {
+		curName, curRngSt, curRngEnd, err := sriovnetworkv1.ParsePFName(curPf)
+		if err != nil {
+			return false, fmt.Errorf("invalid PF name: %s", curPf)
+		}
+		for _, prePf := range previous.Spec.NicSelector.PfNames {
+			preName, preRngSt, preRngEnd, err := sriovnetworkv1.ParsePFName(prePf)
+			if err != nil {
```
pliurh, line 202: We don't need to handle the err for existing policies, right?

zshi-redhat: It loops over all existing policies (exist in kube API) except the policy that has the same name as the policy being evaluated.

pliurh: I mean the err would not happen here. Since all these policies have been verified when it is created. Therefore we can just remove the error handling.

zshi-redhat: Yes, you're right, removed the error checking for previous policy.

PR #281, opened 2020-07-06
unify NIC mapping source for operator, webhook and config daemon

zshi-redhat:


pkg/apis/sriovnetwork/v1/helper.go
```diff
@@ -26,26 +26,74 @@ const (
...
-	"10fb": "10ed",
-	"1015": "1016",
-	"1017": "1018",
+var log = logf.Log.WithName("sriovnetwork")
+var VfIds = []string{}
+
+// NicIdMap contains supported mapping of IDs with each in the format of:
+// Vendor ID, Physical Function Device ID, Virtual Function Device ID
+var NicIdMap = []string{
+	"8086 1583 154c",
```
pliurh, line 35: We may need to remove the NIC models which we don't officially support. Since this PR will also make the webhook to use the same list for admission.

zshi-redhat: removed

PR #284, opened 2020-07-07
Unlimit max mtu size

zshi-redhat:


pliurh:
Maybe we shall just remove the upper limit for MTU.

pliurh:
We need a BZ to backport this change to 4.4.
/lgtm

PR #289, opened 2020-07-09
Add discovery mode to resource injector test

mmirecki:


test/conformance/tests/sriov_operator.go
```diff
@@ -995,50 +995,90 @@ var _ = Describe("[sriov] operator", func() {
...
+							sriovInfos, operatorNamespace,
 
-						Spec: sriovv1.SriovNetworkNodePolicySpec{
-							NodeSelector: map[string]string{
-								"kubernetes.io/hostname": node,
+							func(policy sriovv1.SriovNetworkNodePolicy) bool {
+								if !defaultFilterPolicy(policy) {
+									return false
+								}
+								if &policy.Spec.NicSelector == nil || &policy.Spec.NicSelector.PfNames == nil || len(policy.Spec.NicSelector.PfNames) != 1 {
```
fedepaol, line 1014: Why do we need the check of the nicselector (and why we did not need it before in the generic policy case) ?
We should accept policies with other types of selector (I think pfNames is the less likely to be used)

test/util/discovery/discovery.go
```diff
@@ -41,7 +41,7 @@ func DiscoveredResources(clients *client.ClientSet, sriovInfos *cluster.EnabledN
 			if err != nil {
 				continue
 			}
-			device, ok := filterDevices(node.ObjectMeta.Name, sriovDeviceList)
+			device, ok := filterDevices(policy, node.ObjectMeta.Name, sriovDeviceList)
```
fedepaol, line 44: if the goal is to find the device associated with the resourcename, I think a better way is to pass via the nodestate spec:

```
  spec:
    dpConfigVersion: "1607381"
    interfaces:
    - name: ens1f0
      numVfs: 5
      pciAddress: "0000:12:00.0"
      vfGroups:
      - deviceType: netdevice
        policyName: test-sctppolicyq7xbg
        resourceName: sctptestres
        vfRange: 0-4
```

Here you don't depend on the selector of the policy, but you have the information you need to match the resource with the device.

test/conformance/tests/sriov_operator.go
mmirecki: This was a prerequisite for finding a specific nic. Removed, as this is not really needed in this test.

test/util/discovery/discovery.go
mmirecki: This test does not really need finding the specific nic. Removing this check for now. This will reappear in a following PR (using the nodestate spec)

test/conformance/tests/sriov_operator.go
fedepaol: I think this whole test can go under the generic config category, I am not seeing any custom configuration here.

mmirecki:
The test was moved to shared policy tests

PR #297, opened 2020-07-13
Discovery mode for mtu tests

mmirecki:


test/conformance/tests/sriov_operator.go
```diff
@@ -1093,43 +1093,86 @@ var _ = Describe("[sriov] operator", func() {
...
-							Namespace:    operatorNamespace,
-						},
+					if discovery.Enabled() {
+						node, resourceName, numVfs, intf, err = discovery.DiscoveredResources(clients,
+							sriovInfos, operatorNamespace,
+							func(policy sriovv1.SriovNetworkNodePolicy) bool {
+								if !defaultFilterPolicy(policy) {
+									return false
+								}
+								if policy.Spec.Mtu != 9000 {
```
fedepaol, line 1109: Does Mtu need to be EXACTLY 9000?

test/conformance/tests/sriov_operator.go
```diff
@@ -1093,43 +1093,86 @@ var _ = Describe("[sriov] operator", func() {
...
+									return false
+								}
+								return true
+							},
+							func(node string, sriovDeviceList []*sriovv1.InterfaceExt) (*sriovv1.InterfaceExt, bool) {
```
fedepaol, line 1114: Can this just be "filterDevice" as a function that accepts a single device and returns true if we want it?
I think it would be much more readable.

test/util/cluster/cluster.go
```diff
@@ -87,9 +87,10 @@ func (n *EnabledNodes) FindSriovDevices(node string) ([]*sriovv1.InterfaceExt, e
 	if !ok {
 		return nil, fmt.Errorf("Node %s not found", node)
 	}
-	for _, itf := range s.Status.Interfaces {
+
+	for i, itf := range s.Status.Interfaces {
 		if IsDriverSupported(itf.Driver) {
-			devices = append(devices, &itf)
+			devices = append(devices, &s.Status.Interfaces[i])
```
fedepaol, line 93: Thanks for fixing this!

test/util/namespaces/namespaces.go
```diff
@@ -135,14 +135,16 @@ func waitForSriovNetworkDeletion(operatorNamespace string, cs *testclient.Client
...
-func Clean(operatorNamespace, namespace string, cs *testclient.ClientSet) error {
+func Clean(operatorNamespace, namespace string, cs *testclient.ClientSet, discoveryEnabled bool) error {
 	err := CleanPods(namespace, cs)
 	if err != nil {
 		return err
 	}
-	err = CleanPolicies(operatorNamespace, cs)
-	if err != nil {
-		return err
+	if !discoveryEnabled {
```
fedepaol, line 143: Can you use the global method here `discovery.Enabled` ?

test/conformance/tests/sriov_operator.go
mmirecki: No
Afair, a jumbo frame is anything above 1500. Changing to > 1500

test/conformance/tests/sriov_operator.go
mmirecki: I would prefer not to.
This would require a change for other cases we use it, for example to find the unused devices. That is done by creating a pod and checking all the devices on the single pod. If we modify it, we would have to test each device one by one, with a new pod for each device. This might make a real difference in the test times.

test/util/namespaces/namespaces.go
mmirecki: Not really, that would cause cyclic dependecies
discovery uses cluster
cluster uses pod
pod uses namespaces

test/conformance/tests/sriov_operator.go
fedepaol: Agree. Leave as it is

test/util/namespaces/namespaces.go
fedepaol: sounds good then.

PR #311, opened 2020-07-24
[release-4.5] Bug 1860286: load plugins only when none has been loaded

pliurh:


zshi-redhat:
why this fixes bz1860286, could you explain the reason?

pliurh:
bz1860286 happened is because the plugins were reloaded when the nodeStateAddHandler was triggered in an abnormal situation. Normally `nodeStateAddHandler` shall only be triggered during config daemon boot up, after the status of nodestate CR has been initialized by the writer. However, if the user somehow deleted the nodestate CR accidentally, it will be recreated immediately by the controller. In the configure daemon, the `nodeStateAddHandler` will also be invoked right after that, before the writer gets a chance to initialize the status. In consequence, no plugin except the general plugin can be loaded. So by this change, we don't reload the plugins, we keep using what has been loaded during the bootup. It follows the same logic as we have in 4.6 code.

zshi-redhat:
/lgtm

> bz1860286 happened is because the plugins were reloaded when the nodeStateAddHandler was triggered in an abnormal situation. Normally `nodeStateAddHandler` shall only be triggered during config daemon boot up, after the status of nodestate CR has been initialized by the writer. However, if the user somehow deleted the nodestate CR accidentally, it will be recreated immediately by the controller. In the configure daemon, the `nodeStateAddHandler` will also be invoked right after that, before the writer gets a chance to initialize the status. In consequence, no plugin except the general plugin can be loaded. So by this change, we don't reload the plugins, we keep using what has been loaded during the bootup. It follows the same logic as we have in 4.6 code.

ack
/lgtm

PR #313, opened 2020-07-27
Watch the node state CR by controller

pliurh:


zshi-redhat:
what's the problem to solve?

pliurh:
> what's the problem to solve?

The node state CR cannot be recreated by the controller if it's deleted by the user accidentally.

PR #314, opened 2020-07-28
BUG 1853785: Retry for VF link to become ready

zshi-redhat:
VF link may not be immediately ready after VF creation.

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -19,6 +19,7 @@ import (
 	"github.com/jaypipes/ghw"
 	sriovnetworkv1 "github.com/openshift/sriov-network-operator/pkg/apis/sriovnetwork/v1"
 	"github.com/vishvananda/netlink"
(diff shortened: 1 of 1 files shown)


pkg/utils/utils.go
```diff
@@ -460,10 +474,15 @@ func setVfsAdminMac(iface *sriovnetworkv1.InterfaceExt) error {
 			glog.Errorf("setVfsAdminMac(): unable to get VF id %+v %q", iface.PciAddress, err)
 			return err
 		}
+		_, err = vfIsReady(addr)
+		if err != nil {
+			glog.Errorf("setVfsAdminMac(): VF link is not ready for device %+v %q", addr, err)
+			return err
+		}
 		vfName := tryGetInterfaceName(addr)
```
pliurh, line 482: As we've got the vf name in `vfIsReady`, can we remove this line?  And also for the `vfLink` line bellow.

zshi-redhat: fixed

PR #315, opened 2020-07-29
Add Mellanox CX-4 and CX-6 ib NIC to supported nic list

pliurh:


pkg/apis/sriovnetwork/v1/helper.go
```diff
@@ -35,6 +35,7 @@ var NicIdMap = []string{
 	"8086 158b 154c",
 	"15b3 1015 1016",
 	"15b3 1017 1018",
+	"15b3 101b 101c",
```
zshi-redhat, line 38: CX-4: `15b3 1013 1014`

pliurh: Done

PR #317, opened 2020-07-29
Set logLevel to 2 when recreating default SriovOperatorConfig

zshi-redhat:


SchSeba:
we have the creation when the operator starts here https://github.com/openshift/sriov-network-operator/blob/master/cmd/manager/main.go#L148

This change is only in case someone remove the `default` object right?

zshi-redhat:
> we have the creation when the operator starts here https://github.com/openshift/sriov-network-operator/blob/master/cmd/manager/main.go#L148
> 
> This change is only in case someone remove the `default` object right?

yes, just make sure the re-created sriovOperatorConfig has the right value for LogLevel

PR #320, opened 2020-07-30
Fix nil pointer exception in MTU connectivity test.

fedepaol:


fedepaol:
/override ci/prow/e2e-aws 
The changed test is not even executed

PR #323, opened 2020-08-10
BUG 1867529: mellanox plugin - fix resetting unsupported NICs

zshi-redhat:


Mmduh-483:
@zshi-redhat Did you tested the fix with mellanox nic?

zshi-redhat:
> @zshi-redhat Did you test the fix with mellanox nic?

yes

zshi-redhat:
@pliurh this is likely the issue you hit that switchdev config be reverted automatically.

adrianchiris:
looks good to me.

PR #328, opened 2020-08-21
Bug 1876822: Workaround for unexpected EOF during node state lookup

mmirecki:
The PR is a temporary fix for an issue experienced in test runs.
The issue occurs very sporadically and as such is difficult to debug properly.
It seems that once every few hundred runs an EOF error is returned without a valid reason (or so it seems).

### DIFF
--- a/test/conformance/tests/sriov_operator.go
+++ b/test/conformance/tests/sriov_operator.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"io"
 	"os"
 	"strconv"
 	"strings"
@@ -1701,8 +1702,19 @@ func waitForSRIOVStable() {
 	// TODO: find a better way to handle this scenario
 
 	time.Sleep(5 * time.Second)
+
+	eofErrorCount := 0
 	Eventually(func() bool {
 		res, err := cluster.SriovStable(operatorNamespace, clients)
+		// The check for EofError is done to temorarily work around an issue
+		// occuring during tests run. The issue occurs very sporadicly and as such
+		// is difficult to identify. eofErrorCount is introduced to allow us to respond
+		// to real issues.
+		if err == io.ErrUnexpectedEOF {
+			eofErrorCount++
+			Expect(eofErrorCount).To(BeNumerically("<", 2))
+			return false
+		}
 		Expect(err).ToNot(HaveOccurred())
 		return res
 	}, waitingTime, 1*time.Second).Should(BeTrue())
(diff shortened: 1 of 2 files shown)


test/util/cluster/cluster.go
```diff
@@ -115,8 +116,13 @@ func (n *EnabledNodes) FindOneMellanoxSriovDevice(node string) (*sriovv1.Interfa
 func SriovStable(operatorNamespace string, clients *testclient.ClientSet) (bool, error) {
 	nodeStates, err := clients.SriovNetworkNodeStates(operatorNamespace).List(context.Background(), metav1.ListOptions{})
 	if err != nil {
+		switch err {
```
fedepaol, line 119: This can be a single switch, without the extra nesting.
```
switch {
	err == o.ErrUnexpectedEOF:
		
	case err != nil:

	}
```

test/util/cluster/cluster.go
```diff
@@ -225,3 +231,11 @@ func IsClusterStable(clients *testclient.ClientSet) (bool, error) {
 
 	return true, nil
 }
+
+func IsEofError(err error) bool {
```
fedepaol, line 235: Why the function and not comparing the error to the value?
I'd say it's more clear and the function does not add information.

test/conformance/tests/sriov_operator.go
```diff
@@ -1701,9 +1701,20 @@ func waitForSRIOVStable() {
 	// TODO: find a better way to handle this scenario
 
 	time.Sleep(5 * time.Second)
+
+	eofErrorCount := 0
 	Eventually(func() bool {
 		res, err := cluster.SriovStable(operatorNamespace, clients)
-		Expect(err).ToNot(HaveOccurred())
+		if err != nil {
```
fedepaol, line 1708: This could be unrolled to avoid the extra nesting.
Ideally, the nested part is where the error is handled.
See https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88

Also, in case of handled error I'd explicitly return false.
For example:
```
        if err == io.ErrUnexpectedEOF {
            eofErrorCount++
            Expect(eofErrorCount). // To be greater than 2
            return false
        }
        
        Expect(err).ToNot(HaveOccurred())
}

fedepaol (commented):
A couple of nits, looks good in general

test/util/cluster/cluster.go
mmirecki: Done

test/util/cluster/cluster.go
mmirecki: Done

test/conformance/tests/sriov_operator.go
mmirecki: Done

fedepaol:
/approve
/lgtm
@mmirecki this needs a bz to get merged.

PR #329, opened 2020-08-24
Bug 1870074: Reset PF device whose name doesn't match policy pfNames.

zshi-redhat:


zshi-redhat:
This won't fix the issue in bug1870074, The real fix of this issue will rely on the rootDevice selector in device plugin, which is a new feature that may be added in 4.7.

PR #337, opened 2020-09-03
Bug 1878588: Add image reference for sriov-operator-must-gather

zshi-redhat:


zshi-redhat:
@sosiouxme I'm adding a must-gather image reference in the csv manifests for this OLM managed Operator (SR-IOV). The image will not be used in the operator, it only serves as a reference for user to get the exact must-gather image link. Would you please help to check if this is the right approach for adding a must-gather image link?

The ocp-build-auto for sriov-network-must-gather is located at: https://github.com/openshift/ocp-build-data/blob/openshift-4.6/images/sriov-network-must-gather.yaml

PR #338, opened 2020-09-03
Bug 1875330: Reset sync error if local status generation is equal remote.

fedepaol:
When the fetch of the state fails, a sync error is written on the status.
Then, on next reconciliation loops, if the local generation equals to the remote one, that error is never reset. 
Here we assume that if the two versions are identical, the sync happened successfully and we reset the message.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -361,6 +361,14 @@ func (dn *Daemon) nodeStateSyncHandler(generation int64) error {
 	latest := latestState.GetGeneration()
 	if dn.nodeState.GetGeneration() == latest {
 		glog.V(0).Infof("nodeStateSyncHandler(): Interface not changed")
+		if latestState.Status.LastSyncError != "" ||
+			latestState.Status.SyncStatus != "Succeeded" {
+			dn.refreshCh <- Message{
+				syncStatus:    "Succeeded",
+				lastSyncError: "",
+			}
+		}
+
 		return nil
 	}
 


fedepaol:
@zshi-redhat @pliurh PTAL

zshi-redhat:
> When the fetch of the state fails, a sync error is written on the status.
> Then, on next reconciliation loops, if the local generation equals to the remote one, that error is never reset.

If the local generation equals to the remote one, it means next reconciliation succeeds, which in turn should set the syncStatus to `Succeeded` [here](https://github.com/openshift/sriov-network-operator/blob/master/pkg/daemon/daemon.go#L454), how can it not be reset?

> Here we assume that if the two versions are identical, the sync happened successfully and we reset the message.

pliurh:
> > When the fetch of the state fails, a sync error is written on the status.
> > Then, on next reconciliation loops, if the local generation equals to the remote one, that error is never reset.
> 
> If the local generation equals to the remote one, it means next reconciliation succeeds, which in turn should set the syncStatus to `Succeeded` [here](https://github.com/openshift/sriov-network-operator/blob/master/pkg/daemon/daemon.go#L454), how can it not be reset?

It means LAST reconciliation succeeded. 
This bug would occur when there is a nodeState fetching failed that happened in periodic reconciliation, the `syncStatus` is set to Failed.  But in the next round, when nodeState can be fetched, the config daemon found there is nothing changed, it will just return without reseting the state.

> 
> > Here we assume that if the two versions are identical, the sync happened successfully and we reset the message.

zshi-redhat:
> > > When the fetch of the state fails, a sync error is written on the status.
> > > Then, on next reconciliation loops, if the local generation equals to the remote one, that error is never reset.
> > 
> > 
> > If the local generation equals to the remote one, it means next reconciliation succeeds, which in turn should set the syncStatus to `Succeeded` [here](https://github.com/openshift/sriov-network-operator/blob/master/pkg/daemon/daemon.go#L454), how can it not be reset?
> 
> It means LAST reconciliation succeeded.

Ack, the change looks good to me.

PR #341, opened 2020-09-10
Add webhook delete operation

zshi-redhat:
There are default sriovnetworknodepolicy and sriovoperatorconfig CRs
which we don't want user to delete although they can be re-created.
This PR adds support for DELETE operation in operator-webhook to
block the deletion of default node policy and operator config CRs.


zshi-redhat:
/hold
until master is open to 4.7

PR #344, opened 2020-09-16
Bug 1878565: merge device plugin configs that share the same resource name

zshi-redhat:


pkg/controller/sriovnetworknodepolicy/sriovnetworknodepolicy_controller.go
```diff
@@ -570,8 +570,12 @@ func renderDevicePluginConfigData(pl *sriovnetworkv1.SriovNetworkNodePolicyList)
 		}
 
 		found, i := resourceNameInList(p.Spec.ResourceName, &rcl)
-		netDeviceSelectors := dptypes.NetDeviceSelectors{}
```
pliurh, line 573: I think we can keep this unchanged, and not declare later for each condition, can't we?

zshi-redhat: done

PR #352, opened 2020-09-21
Support RDMA namespace changing and generate CNI config with RDMA isolation support

Mmduh-483:


Mmduh-483:
@zshi-redhat @pliurh Can you please review the PR

bindata/manifests/cni-config/sriov-cni-config.yaml
```diff
@@ -33,6 +34,14 @@ spec:
...
   "link_state":"{{.SriovCniState}}",
+{{- end -}}
+{{- if and (.RdmaIsolation) (eq .CniType "ib-sriov") -}}
+  "rdmaIsolation": true,
 {{- end -}}
   {{.SriovCniIpam}}
-}'
+{{- if and (.RdmaIsolation) (eq .CniType "sriov") -}}
+  },
+  {"type": "rdma"
```
pliurh, line 44: @zshi-redhat will we ship rdma-cni in OCP?

bindata/scripts/add-rdma-service.sh
```diff
@@ -0,0 +1,39 @@
...
+    exit 0
+  fi
+  sleep $WAIT_TIME
+done' > /etc/init.d/sriov-network-rdma.sh
+
+rdma_mode=$(rdma system | cut -c7- | grep "[a-z]")
+
+sed -i 's/new_rdma_mode=.*/new_rdma_mode='$rdma_mode'/g' /etc/init.d/sriov-network-rdma.sh
+
+cat > /usr/lib/systemd/system/sriov-network-operator-rdma.service <<EOF
```
pliurh, line 27: Do we have to use a systemd service to change the rdma mode? And in Openshift, we prefer to use MCO to manage all the systemd services.

pliurh (commented):
I'm working on a refactor patch for upgrading the operator-sdk. I prefer we revisit this PR until that is finished.

bindata/manifests/cni-config/sriov-cni-config.yaml
zshi-redhat: not planned in 4.7.

bindata/scripts/add-rdma-service.sh
Mmduh-483: @pliurh yes, changing rdma mode fails when there are containers running
For openshift MCO it can be in a follow-up PR

PR #356, opened 2020-09-24
Bug 1880990: SR-IOV conformance flakes.

fedepaol:
- Tests starting when devices are not available yet. Waiting for them to
appear under the node.

- A macvlan flake complaining about device not ready, sticking with default.

### DIFF
--- a/test/util/network/network.go
+++ b/test/util/network/network.go
@@ -100,7 +100,7 @@ func GetSriovNicIPs(pod *k8sv1.Pod, ifcName string) ([]string, error) {
 // Return a definition of a macvlan NetworkAttachmentDefinition
 // name name
 // namespace namespace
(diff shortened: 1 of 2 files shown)


test/util/network/network.go
```diff
@@ -100,7 +100,7 @@ func GetSriovNicIPs(pod *k8sv1.Pod, ifcName string) ([]string, error) {
 // Return a definition of a macvlan NetworkAttachmentDefinition
 // name name
 // namespace namespace
-func CreateMacvlanNetworkAttachmentDefinition(name string, namespace string, nodeNicName string) netattdefv1.NetworkAttachmentDefinition {
+func CreateMacvlanNetworkAttachmentDefinition(name string, namespace string) netattdefv1.NetworkAttachmentDefinition {
```
SchSeba, line 103: please add in the comment here that because we are not give him any interface the host will select one, pointing to the doc please

PR #357, opened 2020-09-25
Bug 1880990: Fix PF shutdown test. On some nodes route link returns void.

fedepaol:
Here we use another approach to see if the interface is slave of the bridge.
Also, this test proved itself to be pretty dangerous, disabling it in
discovery mode.

Signed-off-by: Federico Paolinelli <fpaoline@redhat.com>


fedepaol:
/hold
requires validation

SchSeba (commented):
/lgtm

/hold

please remove the hold after validation the code looks good!

PR #358, opened 2020-09-25
Bug 1880990: Fix syntax error in conformance test.

fedepaol:
While removing from discovery mode, a syntax error was introduced.


fromanirh:
/lgtm
since reviewing this PR seems to require only golang knowledge (not sriov-specific :) )

PR #364, opened 2020-10-02
Bug 1884251: Conformance test: make sure that the hostnetwork pod and the test pod run on the same node.

fedepaol:
In regular mode, this is ensured by the fact that we configure the devices only on the first node,
and as a side effect both pods get scheduled on the same node. This is not true in discovery mode
and with this change we ensure that.

### DIFF
--- a/test/conformance/tests/sriov_operator.go
+++ b/test/conformance/tests/sriov_operator.go
@@ -264,7 +264,7 @@ var _ = Describe("[sriov] operator", func() {
 			intf := &sriovv1.InterfaceExt{}
 
 			validationFunction := func(networks []string, containsFunc func(line string) bool) {
-				podObj := pod.DefineWithNetworks(networks)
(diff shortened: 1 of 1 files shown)


test/conformance/tests/sriov_operator.go
```diff
@@ -265,6 +265,7 @@ var _ = Describe("[sriov] operator", func() {
 
 			validationFunction := func(networks []string, containsFunc func(line string) bool) {
 				podObj := pod.DefineWithNetworks(networks)
+				podObj = pod.RedefineWithNodeSelector(podObj, node)
```
SchSeba, line 268: nit: can you do something like this?

```golang
podObj := pod.RedefineWithNodeSelector(pod.DefineWithNetworks(networks), node)
```

fedepaol: Done

SchSeba (approved):
/lgtm

PR #369, opened 2020-10-07
Add support for running in a virtual machine -- Remove PF dependency

atyronesmith:
This PR adds support for running (initially) in an OpenShift on OpenStack environment.  In a VM environment, only VFs are attached (no PF) and the VFs represent an underlying OpenStack network.  This PR removes the necessity of accessing a PF and uses additional information passed into the VM by the OpenStack metadata service to include underlying network mappings in the sriovnetworknodestate.  In addition, the exposed underlying network information can be used to select networks for a policy.  

This PR is in conjunction with a PR in  k8snetworkplumbingwg/sriov-network-device-plugin repository.  

This PR is to allow a review of the concept.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -41,6 +41,27 @@ import (
 	"github.com/openshift/sriov-network-operator/pkg/utils"
 )
 
+// PlatformType foo
+type PlatformType int
+
+const (
+	// Baremetal platform
+	Baremetal PlatformType = iota
+	// Virtual platform
+	Virtual
+)
+
+func (e PlatformType) String() string {
+	switch e {
+	case Baremetal:
+		return "Baremetal"
+	case Virtual:
+		return "Virtual"
+	default:
+		return fmt.Sprintf("%d", int(e))
+	}
+}
+
 const (
 	// updateDelay is the baseline speed at which we react to changes.  We don't
 	// need to react in milliseconds as any change would involve rebooting the node.
@@ -59,6 +80,8 @@ type Daemon struct {
 	// name is the node name.
 	name      string
 	namespace string
+
+	platform PlatformType
 
 	client snclientset.Interface
 	// kubeClient allows interaction with Kubernetes, including the node we are running on.
@@ -128,9 +151,11 @@ func New(
 	stopCh <-chan struct{},
 	syncCh <-chan struct{},
 	refreshCh chan<- Message,
+	platformType PlatformType,
 ) *Daemon {
 	return &Daemon{
 		name:       nodeName,
+		platform:   platformType,
 		client:     client,
 		kubeClient: kubeClient,
 		exitCh:     exitCh,
@@ -544,8 +569,14 @@ func (dn *Daemon) restartDevicePluginPod() error {
 }
 
 func (dn *Daemon) loadVendorPlugins(ns *sriovnetworkv1.SriovNetworkNodeState) error {
-	pl := registerPlugins(ns)
-	pl = append(pl, GenericPlugin)
+	var pl []string
+
+	if dn.platform == Virtual {
+		pl = append(pl, VirtualPlugin)
+	} else {
+		pl = registerPlugins(ns)
+		pl = append(pl, GenericPlugin)
+	}
 	dn.LoadedPlugins = make(map[string]VendorPlugin)
 
 	for _, pn := range pl {
--- a/pkg/plugins/virtual/virtual_plugin.go
+++ b/pkg/plugins/virtual/virtual_plugin.go
@@ -1,0 +1,141 @@
+package main
+
+import (
+	"os/exec"
+	"reflect"
+	"syscall"
+
+	"github.com/golang/glog"
+	sriovnetworkv1 "github.com/openshift/sriov-network-operator/api/v1"
+	"github.com/openshift/sriov-network-operator/pkg/utils"
+)
+
+// VirtualPlugin Plugin type to use on a virtual platform
+type VirtualPlugin struct {
+	PluginName     string
+	SpecVersion    string
+	DesireState    *sriovnetworkv1.SriovNetworkNodeState
+	LastState      *sriovnetworkv1.SriovNetworkNodeState
+	LoadVfioDriver uint
+}
+
+const (
+	unloaded = iota
+	loading
+	loaded
+)
+
+// Plugin VirtualPlugin type
+var Plugin VirtualPlugin
+
+// Initialize our plugin and set up initial values
+func init() {
+	Plugin = VirtualPlugin{
+		PluginName:     "virtual_plugin",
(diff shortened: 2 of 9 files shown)


zshi-redhat (commented):
Thanks Aaron for working on the patches, added few comments inline. 
I think the overall design looks good to me:
1. A virtual plugin for VM use case
2. Auto detect platform type and invoke the virtual plugin

pkg/utils/utils_virtual.go
```diff
@@ -0,0 +1,312 @@
...
+		if err != nil {
+			glog.Warningf("DiscoverSriovDevicesVirtual(): unable to parse device class for device %+v %q", device, err)
+			continue
+		}
+		if devClass != netClass {
+			// Not network device
+			continue
+		}
+
+		if isVirtio(device.Address) {
```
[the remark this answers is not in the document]
atyronesmith: Ok..  Good.  I was unsure about this.

pkg/utils/utils_virtual.go
```diff
@@ -0,0 +1,312 @@
...
+			iface.Name = name
+			iface.Mac = getNetDevMac(name)
+			iface.LinkSpeed = getNetDevLinkSpeed(name)
+		}
+		iface.LinkType = getLinkType(iface)
+
+		iface.TotalVfs = 1
+		iface.NumVfs = 1
+
+		vf := sriovnetworkv1.VirtualFunction{
```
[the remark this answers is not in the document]
atyronesmith: I think it would be good to treat VFs as individual devices.  The problem is that the current code seems to be keyed to TotalVFs/NumVfs and I took the approach of setting to 1 to reduce the number code changes.  I can take a look at your suggestion.

pkg/utils/utils.go
```diff
@@ -260,6 +383,53 @@ func configSriovDevice(iface *sriovnetworkv1.Interface, ifaceStatus *sriovnetwor
...
+	glog.V(2).Infof("configSriovDeviceVirtual(): config interface %s with %v", iface.PciAddress, iface)
+	// Config VFs
+	if iface.NumVfs > 0 {
+		if iface.NumVfs > 1 {
+			glog.Warningf("configSriovDeviceVirtual(): in a virtual environment, only one VF per interface (NumVfs: %d)", iface.NumVfs)
+			return errors.New("NumVfs > 1")
+		}
+		if len(iface.VfGroups) != 1 {
+			glog.Warningf("configSriovDeviceVirtual(): missing VFGroup")
+			return errors.New("NumVfs > 1")
```
martinkennelly: I think this should be VfGroups != 1 error message instead of NumVfs > 1.

pkg/daemon/daemon.go
```diff
@@ -486,6 +511,11 @@ func (dn *Daemon) completeDrain() error {
 	return nil
 }
 
+func (dn *Daemon) discoverPlatform() error {
```
zshi-redhat, line 514: dn.platform is discovered and passed by operator controller when daemon.New is called, do we need this function in daemon?

pkg/daemon/daemon.go
```diff
@@ -544,8 +574,14 @@ func (dn *Daemon) restartDevicePluginPod() error {
 }
 
 func (dn *Daemon) loadVendorPlugins(ns *sriovnetworkv1.SriovNetworkNodeState) error {
-	pl := registerPlugins(ns)
-	pl = append(pl, GenericPlugin)
+	var pl []string
```
zshi-redhat, line 577: I think we can re-use the same registerPlugins functions for both virtual and baremetal by moving this piece of code into registerPlugins and pass an additional argument to it. for example:

```
registerPlugins(ns, dn.platform)
```

zshi-redhat (commented):
Looks generally good to me.
I think we might want to do two things in a follow-up PRs:
1) enable use of numVfs=0 so that virtual-plugin doesn't need to populate `InterfaceExt.VFs`
2) move platform specific code into a separate file.

pkg/daemon/daemon.go
atyronesmith: The problem is that with virtual, we don't want to load the generic plugin.

controllers/sriovnetworknodepolicy_controller.go
```diff
@@ -545,6 +545,10 @@ func renderDevicePluginConfigData(pl *sriovnetworkv1.SriovNetworkNodePolicyList,
 				netDeviceSelectors.Drivers = sriovnetworkv1.UniqueAppend(netDeviceSelectors.Drivers, p.Spec.DeviceType)
 			}
 
+			if len(p.Spec.NicSelector.RootDevices) > 0 {
```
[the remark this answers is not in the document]
atyronesmith: It is... will fix.

pkg/daemon/daemon.go
atyronesmith: We do not and I thought I had removed it...

pkg/daemon/daemon.go
zshi-redhat: Yes, you're right, I forgot the generic-plugin is not needed in virtual env.

pkg/plugins/virtual/virtual_plugin.go
```diff
@@ -0,0 +1,141 @@
...
+func (p *VirtualPlugin) OnNodeStateChange(old, new *sriovnetworkv1.SriovNetworkNodeState) (needDrain bool, needReboot bool, err error) {
+	glog.Info("virtual-plugin OnNodeStateChange()")
+	needDrain = false
+	needReboot = false
+	err = nil
+	p.DesireState = new
+
+	if p.LoadVfioDriver != loaded {
+		if needVfioDriver(new) {
+			p.LoadVfioDriver = loading
```
zshi-redhat, line 78: @atyronesmith I noticed the iommu kernel args change was removed, is it handled by openstack ?

pkg/utils/utils_virtual.go
```diff
@@ -0,0 +1,179 @@
...
+			glog.Warningf("configSriovDeviceVirtual(): in a virtual environment, only one VF per interface (NumVfs: %d)", iface.NumVfs)
+			return errors.New("NumVfs > 1")
+		}
+		if len(iface.VfGroups) != 1 {
+			glog.Warningf("configSriovDeviceVirtual(): missing VFGroup")
+			return errors.New("NumVfs != 1")
+		}
+		vfAddrs := []string{iface.PciAddress}
+		glog.V(2).Infof("configSriovDeviceVirtual(): vfaddrs %v", vfAddrs)
+		for _, addr := range vfAddrs {
```
pliurh, line 148: Can `len(vfAddrs)` be larger than 1? Why we need a loop here?

atyronesmith (commented):
Completed

PR #370, opened 2020-10-10
Bug 1877648: Support generating device plugin configmap with rootDevice selector

zshi-redhat:


zshi-redhat:
Rebase against master branch

PR #373, opened 2020-10-16
Create manifests for 4.7

pliurh:


zshi-redhat:
Why the old version manifests are removed from master?

pliurh:
> Why the old version manifests are removed from master?

The old version folders are only consumed by ART on the specific branch.

zshi-redhat:
> > Why the old version manifests are removed from master?
> 
> The old version folders are only consumed by ART on the specific branch.

In case there is a bug in the manifests for released branch, do we need to merge the fix in master branch, then backport all the way to the specific branch?

pliurh:
> > > Why the old version manifests are removed from master?
> > 
> > 
> > The old version folders are only consumed by ART on the specific branch.
> 
> In case there is a bug in the manifests for released branch, do we need to merge the fix in master branch, then backport all the way to the specific branch?

I don't think we need to.

PR #376, opened 2020-10-21
Bug 1890088: Conformance tests: Wait for the resources to be ready.

fedepaol:
There are a couple of tests where we don't wait for the resource number to be ready.
This makes the tests flakes as the pod waiting time must take also into account the time
needed for having the resources to be ready.


fedepaol:
/hold
still needs validation

PR #372, opened 2020-10-16
Update operator-sdk to 1.0.1

pliurh:

PR #378, opened 2020-10-23
Add an optional SriovNetwork section: additional metaplugins

fedepaol:
In order to generate a network attachment definition that chains the SR-IOV CNI configuration with additional metaplugins (such as VRF, or firewall), we expose a new MetaPluginsConfig where to put the json list of the extra configurations, in the form of

```
{
	// meta plugin A config
},
{
	// meta plugin B config
}
```

### DIFF
--- a/api/v1/sriovnetwork_types.go
+++ b/api/v1/sriovnetwork_types.go
@@ -42,6 +42,9 @@ type SriovNetworkSpec struct {
 	// +kubebuilder:validation:Minimum=0
 	// Maximum tx rate, in Mbps, for the VF. Defaults to 0 (no rate limiting)
 	MaxTxRate *int `json:"maxTxRate,omitempty"`
+	// MetaPluginsConfig configuration to be used in order to chain metaplugins to the sriov interface returned
+	// by the operator.
+	MetaPluginsConfig string `json:"metaplugins,omitempty"`
 }
 
 // SriovNetworkStatus defines the observed state of SriovNetwork
--- a/bindata/manifests/cni-config/sriov-cni-config.yaml
+++ b/bindata/manifests/cni-config/sriov-cni-config.yaml
@@ -9,6 +9,10 @@ spec:
   config: '{
   "cniVersion":"0.3.1",
   "name":"{{.SriovNetworkName}}",
+{{- if .MetaPluginsConfigured -}}
+  "plugins": [
+    {
+{{- end -}}
   "type":"{{.CniType}}",
 {{- if eq .CniType "sriov" -}}
   "vlan":{{.SriovCniVlan}},
@@ -35,4 +39,11 @@ spec:
   "link_state":"{{.SriovCniState}}",
 {{- end -}}
   {{.SriovCniIpam}}
-}'
+}
+{{- if .MetaPluginsConfigured -}}
+  ,
+  {{.MetaPlugins}}
+  ]
+}
+{{- end -}}
+'
--- a/README.md
+++ b/README.md
@@ -58,6 +58,45 @@ spec:
   resourceName: intelnics
 ```
 
+#### Chaining CNI metaplugins
+
+It is possible to add additional capabilities to the device configured via the SR-IOV configuring optional metaplugins.
+
+In order to do this, the `metaPlugins` field must contain the array of one or more additional configurations used to build a [network configuration list](https://github.com/containernetworking/cni/blob/master/SPEC.md#network-configuration-lists), as per the following example:
+
+```yaml
+apiVersion: sriovnetwork.openshift.io/v1
+kind: SriovNetwork
+metadata:
+  name: example-network
+  namespace: example-namespace
+spec:
+  ipam: |
+    {
+      "type": "host-local",
+      "subnet": "10.56.217.0/24",
+      "rangeStart": "10.56.217.171",
+      "rangeEnd": "10.56.217.181",
+      "routes": [{
+        "dst": "0.0.0.0/0"
+      }],
+      "gateway": "10.56.217.1"
+    }
+  vlan: 0
+  resourceName: intelnics
+  metaPlugins : |
+    {
+      "type": "tuning",
+      "sysctl": {
+        "net.core.somaxconn": "500"
+      }
+    },
+    {
+      "type": "vrf",
+      "vrfname": "red"
+    }
+```
+
 ### SriovNetworkNodeState
 
 The custom resource to represent the SR-IOV interface states of each host, which should only be managed by the operator itself.
--- a/api/v1/helper.go
+++ b/api/v1/helper.go
@@ -19,13 +19,13 @@ import (
 )
 
 const (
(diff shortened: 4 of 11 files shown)


fedepaol:
@zshi-redhat @pliurh please take a look.
If the approach sounds sane, I'll add e2e tests and docs.

bindata/manifests/cni-config/sriov-cni-config.yaml
```diff
@@ -9,6 +9,10 @@ spec:
   config: '{
   "cniVersion":"0.3.1",
   "name":"{{.SriovNetworkName}}",
+{{- if .MetaPluginsConfigured -}}
+  "plugins": [
+    {
+  {{- end -}}
```
zshi-redhat, line 15: nit: indent doesn't match with if?

bindata/manifests/cni-config/sriov-cni-config.yaml
```diff
@@ -35,4 +39,10 @@ spec:
   "link_state":"{{.SriovCniState}}",
 {{- end -}}
   {{.SriovCniIpam}}
-}'
+}
+  {{- if .MetaPluginsConfigured -}}
+  ,
+  {{.MetaPlugins}}
```
zshi-redhat, line 45: Do we need curly braces for `{{.MetaPlugins}}`?

zshi-redhat (commented):
This PR may require rebase after https://github.com/openshift/sriov-network-operator/pull/372 is in.

bindata/manifests/cni-config/sriov-cni-config.yaml
fedepaol: Done

bindata/manifests/cni-config/sriov-cni-config.yaml
fedepaol: I guess no:
The idea is to ask the users to fill the array, meaning the expectation is to have them to add the braces (and it's the only way to make it work if they want to chain more than one plugin):
```
{
	// meta plugin A config
},
{
	// meta plugin B config
}
```

zshi-redhat: ack, this is different from how `ipam` is configured, we might need document the string format.

fedepaol: Yes, wanted to make sure the approach looked fine first.
Will change this PR with the docs.

fedepaol:
@zshi-redhat did a few changes, mostly driven by the failing test, such as a missing bracket, and the need for 0.4.0 version for the list.

zshi-redhat:
looks good to me.

api/v1/sriovnetwork_types.go
```diff
@@ -42,6 +42,9 @@ type SriovNetworkSpec struct {
 	// +kubebuilder:validation:Minimum=0
 	// Maximum tx rate, in Mbps, for the VF. Defaults to 0 (no rate limiting)
 	MaxTxRate *int `json:"maxTxRate,omitempty"`
+	// MetaPluginsConfig configuration to be used in order to chain metaplugins to the sriov interface returned
+	// by the operator.
+	MetaPluginsConfig string `json:"metaplugins,omitempty"`
```
pliurh, line 47: Shall we use `runtime.RawExtension` to store the MetaPluginsConfig? It will be useful if we need to manipulate the content of the meta plugin config.

bindata/manifests/cni-config/sriov-cni-config.yaml
```diff
@@ -7,8 +7,16 @@ metadata:
     k8s.v1.cni.cncf.io/resourceName: {{.SriovCniResourceName}}
 spec:
   config: '{
+{{- if .MetaPluginsConfigured -}}
+  "cniVersion":"0.4.0",
```
pliurh, line 11: @zshi-redhat Does sriov-cni support cni spec 0.4.0? If yes, shall we always use version 0.4.0 here?

api/v1/sriovnetwork_types.go
fedepaol: Good point. I used string to be consistent with the ipam field, but I agree makes more sense to use RawExtension.
Will change

pliurh: I want to change the `ipam` field to `runtime.RawExtension` as well. But I am not quite sure whether it will break the backward compatibility or not. We can address that in separate PR later.

fedepaol: Question: if we go with the RawExtension, the user will need to put the yaml-ified version of the cni config, right?

For example 

```
metaplugins: |
    { "type": "firewall", "backend": "iptables" }
```

becomes something like

```
  metaplugins:
    plugins:
      - type: firewall
        backend: iptables

```

Am I missing something? If this is the case, I feel this is a bit worse for the user, as all the meta cni plugin examples are in the json form and they'll need to do the mental translation.
Thoughts?

pliurh: Yeah. That is a problem. I agree it's better to keep using string here.

bindata/manifests/cni-config/sriov-cni-config.yaml
zshi-redhat: I think the major change of 0.4.0 spec is the cmdCheck, we have a dummy interface interface for cmdCheck in sriov-cni.
maybe we shall wait until 0.4.0 is fully supported in sriov-cni.

api/v1/sriovnetwork_types.go
zshi-redhat: yes, use string, or is rawMessage an option?

bindata/manifests/cni-config/sriov-cni-config.yaml
fedepaol: The 0.4.0 version is mandatory for using a list of plugins though.
@zshi-redhat is it fine to leave it as it is (setting only when a meta plugin is added)?

api/v1/sriovnetwork_types.go
fedepaol: I think rawMessage suffers of the same issue: we must declare the field as object, which eventually gets marshalled.
So in that case the user will have to convert the yaml to json.

If we want to leave the json format, I think `string` is the only option.

bindata/manifests/cni-config/sriov-cni-config.yaml
zshi-redhat: Can we set cniVersion for meta plugin only?
Just checked sriov-cni, it accepts all CNI versions as supported, I think we shall be fine running sriov-cni with version 0.4.0 although I didn't test myself. 
Is 0.4.0 the only version supported by VRF? I see in the README page, it uses 0.3.1 as an example, but in the code, it has 0.4.0.

fedepaol:
@zshi-redhat @pliurh I reverted the CNI version to 0.3.1, changed the test to use the tuning plugin which is 0.3.1 compatible.

Please take another look

PR #379, opened 2020-10-26
Add flag to enable SwitchDev mode in SriovNetworkNodePolicy CRD

pliurh:
The generic plugin will skip configuring swichdev devices. It shall be done
through a systemd services on the host.


pliurh:
@moshe010 @Mmduh-483 @adrianchiris PTAL.

pkg/utils/utils.go
```diff
@@ -115,6 +118,10 @@ func SyncNodeState(newState *sriovnetworkv1.SriovNetworkNodeState) error {
 		for _, iface := range newState.Spec.Interfaces {
 			if iface.PciAddress == ifaceStatus.PciAddress {
 				configured = true
+				if iface.SwitchMode == sriovnetworkv1.SWITCHMODE_SWITCHDEV {
+					// Skip for switchdev devices
+					continue
```
zshi-redhat, line 123: nit: break

pkg/plugins/generic/generic_plugin.go
```diff
@@ -186,6 +186,10 @@ func needDrainNode(desired sriovnetworkv1.Interfaces, current sriovnetworkv1.Int
 		for _, iface := range desired {
 			if iface.PciAddress == ifaceStatus.PciAddress {
 				configured = true
+				if iface.SwitchMode == sriovnetworkv1.DEVMODE_SWITCHDEV {
+					// Skip for switchdev devices
+					continue
```
zshi-redhat, line 191: nit: break

pkg/utils/utils.go
```diff
@@ -115,6 +118,10 @@ func SyncNodeState(newState *sriovnetworkv1.SriovNetworkNodeState) error {
 		for _, iface := range newState.Spec.Interfaces {
 			if iface.PciAddress == ifaceStatus.PciAddress {
 				configured = true
+				if iface.SwitchMode == sriovnetworkv1.SWITCHMODE_SWITCHDEV {
+					// Skip for switchdev devices
```
zshi-redhat, line 122: Add a debug log here for skipping switchdev mode device.

pkg/plugins/generic/generic_plugin.go
zshi-redhat, line 191: Add a debug log.

api/v1/helper.go
```diff
@@ -22,6 +22,8 @@ const (
 	MANIFESTS_PATH       = "./bindata/manifests/cni-config"
 	LASTNETWORKNAMESPACE = "operator.sriovnetwork.openshift.io/last-network-namespace"
 	FINALIZERNAME        = "netattdef.finalizers.sriovnetwork.openshift.io"
+	SWITCHMODE_LEGACY    = "legacy"
```
zshi-redhat, line 25: SWITCHMODE - > ESWITCHMODE ?

api/v1/sriovnetworknodepolicy_types.go
```diff
@@ -35,6 +35,9 @@ type SriovNetworkNodePolicySpec struct {
 	// +kubebuilder:validation:Enum=eth;ETH;ib;IB
 	// NIC Link Type. Allowed value "eth", "ETH", "ib", and "IB".
 	LinkType string `json:"linkType,omitempty"`
+	// +kubebuilder:validation:Enum=legacy;switchdev
+	// NIC Device Mode. Allowed value "legacy","switchdev".
+	SwitchMode string `json:"switchMode,omitempty"`
```
zshi-redhat, line 40: SwitchMode -> EswitchMode ?

api/v1/sriovnetworknodestate_types.go
```diff
@@ -20,6 +20,7 @@ type Interface struct {
 	Mtu        int       `json:"mtu,omitempty"`
 	Name       string    `json:"name,omitempty"`
 	LinkType   string    `json:"linkType,omitempty"`
+	SwitchMode string    `json:"switchMode,omitempty"`
```
zshi-redhat, line 23: EswitchMode?

api/v1/sriovnetworknodestate_types.go
```diff
@@ -43,6 +44,7 @@ type InterfaceExt struct {
 	NumVfs     int               `json:"numVfs,omitempty"`
 	LinkSpeed  string            `json:"linkSpeed,omitempty"`
 	LinkType   string            `json:"linkType,omitempty"`
+	SwitchMode string            `json:"switchMode,omitempty"`
```
zshi-redhat, line 47: EswitchMode

zshi-redhat (commented):
How do we handle nodeState syncStatus for switchdev policy?
Until MCO reboots the node, will syncStatus be succeeded?

pliurh (commented):
> How do we handle nodeState syncStatus for switchdev policy?
> Until MCO reboots the node, will syncStatus be succeeded?

That's a good question. In this patch, the config daemon doesn't handle the syncStatus for switchdev policy. It will just become 'Succeed'. If we make it return a failure, the config daemon would keep retrying. I am wondering whether we have a better way to handle it.

zshi-redhat:
> > How do we handle nodeState syncStatus for switchdev policy?
> > Until MCO reboots the node, will syncStatus be succeeded?
> 
> That's a good question. In this patch, the config daemon doesn't handle the syncStatus for switchdev policy. It will just become 'Succeed'. If we make it return a failure, the config daemon would keep retrying. I am wondering whether we have a better way to handle it.

If cluster nodes are applied with mixed policies (switchdev and non-switchdev), would there be a problem that MCO and sriov config daemon are not synced on which node is being rebooted, which may result in interrupted workloads?

Maybe we need a sync mechenism between Operator applying MachineConfig and config daemon?
Shall we use sriov config daemon to write the device switchdev systemd config?

config/crd/bases/sriovnetwork.openshift.io_sriovibnetworks.yaml
```diff
@@ -1,6 +1,6 @@
 
 ---
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
```
Mmduh-483, line 3: I think this is not related to the switchdev

Mmduh-483: There are several CRD files changes not related to eswitch mode, it seems you are not using the same version of the operator-sdk, IMO it can be in another PR

pkg/plugins/generic/generic_plugin.go
```diff
@@ -186,6 +186,11 @@ func needDrainNode(desired sriovnetworkv1.Interfaces, current sriovnetworkv1.Int
 		for _, iface := range desired {
 			if iface.PciAddress == ifaceStatus.PciAddress {
```
Mmduh-483, line 187: No need to continue the loop after finding the device, maybe adding break at end of if

pkg/utils/utils.go
```diff
@@ -595,3 +603,12 @@ func generateRandomGuid() net.HardwareAddr {
 
 	return guid
 }
+
+func GetNicSriovMode(pciAddress string) (string, error) {
```
Mmduh-483, line 607: @pliurh Will `SetNicSriovMode` be in a follow up?

Makefile
```diff
@@ -39,7 +39,7 @@ BUNDLE_METADATA_OPTS ?= $(BUNDLE_CHANNELS) $(BUNDLE_DEFAULT_CHANNEL)
 # Image URL to use all building/pushing image targets
 IMG ?= controller:latest
 # Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
-CRD_OPTIONS ?= "crd:trivialVersions=true"
+CRD_OPTIONS ?= "crd:crdVersions={v1},trivialVersions=true"
```
adrianchiris, line 42: can the CRD version bump be in a separate commit in this PR ? or a separate PR ?

i.e recreate CRDs with the new operator SDK without switchdev change

api/v1/helper.go
```diff
@@ -19,9 +19,11 @@ import (
 )
 
 const (
-	MANIFESTS_PATH       = "./bindata/manifests/cni-config"
-	LASTNETWORKNAMESPACE = "operator.sriovnetwork.openshift.io/last-network-namespace"
-	FINALIZERNAME        = "netattdef.finalizers.sriovnetwork.openshift.io"
+	MANIFESTS_PATH        = "./bindata/manifests/cni-config"
+	LASTNETWORKNAMESPACE  = "operator.sriovnetwork.openshift.io/last-network-namespace"
+	FINALIZERNAME         = "netattdef.finalizers.sriovnetwork.openshift.io"
+	ESWITCHMODE_LEGACY    = "legacy"
```
adrianchiris, line 25: note to self These should really be constants in netlink package.

adrianchiris (commented):
So i would be happy if the commit message provided more context on:

1. The problem this PR is trying to solve
2. How it attempts to solve it (general outline of the changes)

Also an outline on how this will work in the full solution is appreciated.

adrianchiris:
Adding this for reference. https://github.com/openshift/enhancements/pull/296

adrianchiris:
> Also an outline on how this will work in the full solution is appreciated.
as well as more information on the systemd service, what it will do and who will deploy it.

config/crd/bases/sriovnetwork.openshift.io_sriovibnetworks.yaml
pliurh: Yeah, it can be in a separate PR. Those changes were mistakenly introduced by the previous operator upgrade.

pkg/utils/utils.go
pliurh: No, at least not for openshift. The plan is to use the Machine Config Operator to insert a new systemd service which will configure the VF and set NIC to switchdev mode on the host.

pliurh: In this release, we only use switchdev for OVN offloading use case. And we want users can use VF bond with it. In openshift-sdn, bond is created by NetworkManager configure file inserted by MCO. So we have to configure switchdev before NetworkManager starts on the host.

pkg/plugins/generic/generic_plugin.go
pliurh: fixed

Makefile
pliurh: Addressed by https://github.com/openshift/sriov-network-operator/pull/389

pliurh:
> as well as more information on the systemd service, what it will do and who will deploy it.

PTAL https://github.com/openshift/sriov-network-operator/pull/393

pliurh:
@zshi PTAL

pkg/utils/utils.go
```diff
@@ -115,6 +118,11 @@ func SyncNodeState(newState *sriovnetworkv1.SriovNetworkNodeState) error {
 		for _, iface := range newState.Spec.Interfaces {
 			if iface.PciAddress == ifaceStatus.PciAddress {
 				configured = true
+				if iface.EswitchMode == sriovnetworkv1.ESWITCHMODE_SWITCHDEV {
```
zshi-redhat, line 121: @pliurh  this part will be updated in the following PR, right? e.g. To not skip switchdev device in daemon.

pliurh: removed.

Issue #381, opened 2020-10-27
[doc] quickstart.md needs to be updated


mrVectorz:
The quickstart guide should reflect the latest commit to the Makefile:
- [b21e53957798a8bbdc12b8025b8223e9d9912493](https://github.com/openshift/sriov-network-operator/commit/b21e53957798a8bbdc12b8025b8223e9d9912493)

The section here is no longer required and will only output an error that the rule is missing (which it now is)
```
Install the Operator-SDK. The following commands will put operator-sdk to your $GOPATH/bin, please make sure that path is included in your $PATH.

cd $GOPATH/src/github.com/openshift/sriov-network-operator
make operator-sdk
```
This is the error:
```
[root@openshift-jumpserver-0 sriov-network-operator]# make operator-sdk
which: no controller-gen in (/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/root/bin:/usr/local/go/bin:/root/golang/bin)
which: no kustomize in (/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/root/bin:/usr/local/go/bin:/root/golang/bin)
make: *** No rule to make target 'operator-sdk'.  Stop.
```

PR #382, opened 2020-10-27
Reflect updated Makefile in Quickstart doc

mrVectorz:
Makefile rule `operator-sdk` has been removed

Closes issue #381

### DIFF
--- a/doc/quickstart.md
+++ b/doc/quickstart.md
@@ -26,6 +26,8 @@
 
 ## Installation
 
+Mase sure to have installed the Operator-SDK, as shown in its [install documentation](https://sdk.operatorframework.io/docs/installation/install-operator-sdk/), and that the binaries are available in your \$PATH.
+
 Clone this GitHub repository.
 
 ```bash
--- a/doc/quickstart.md
+++ b/doc/quickstart.md
@@ -26,17 +26,10 @@
 
 ## Installation
 
-Firstly, clone this GitHub repository.
+Clone this GitHub repository.
 
 ```bash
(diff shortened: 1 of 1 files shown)


zshi-redhat:
@pliurh do you have plan to add the operator-sdk entry back in Makefile?

doc/quickstart.md
```diff
@@ -26,19 +26,12 @@
...
 ## Installation
 
-Firstly, clone this GitHub repository.
+Clone this GitHub repository.
 
 ```bash
 go get github.com/openshift/sriov-network-operator
 ```
 
-Install the Operator-SDK. The following commands will put operator-sdk to your $GOPATH/bin, please make sure that path is included in your \$PATH.
```
pliurh, line 35: Instead of removing this paragraph, I think it is better we put the link of operator-sdk installation https://sdk.operatorframework.io/docs/installation/install-operator-sdk/

pliurh:
> @pliurh do you have plan to add the operator-sdk entry back in Makefile?

No.

doc/quickstart.md
mrVectorz: In case they don't have it installed?
`deploy-cluster` will install controller-gen and kustomize if they are missing anyway.
I can of course add it if you see it necessary though.

pliurh: Yes.
We need `operator-sdk` in the `bundle` target. Although we don't use the automatically generated bundle at this moment, we plan to.

mrVectorz: @pliurh Alright, added the operator-sdk install link as requested

PR #388, opened 2020-11-01
Add DisableDrain flag into operator config for efficient debugging

roytman:
SR-IOV Operator drains nodes before applying any configuration on the node NIC. this usually takes some time since it needs to evict the pods on that node. 
This PR adds a new DisableDrain flag in Operator Config CRD to allow disabling of drain process when applying a new policy which in turn saves the waiting time.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -88,6 +88,8 @@ type Daemon struct {
 
 	drainable bool
 
+	disableDrain bool
+
 	nodeLister listerv1.NodeLister
 
 	workqueue workqueue.RateLimitingInterface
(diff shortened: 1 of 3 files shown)


pkg/daemon/daemon.go
```diff
@@ -346,6 +348,11 @@ func (dn *Daemon) operatorConfigChangeHandler(old, new interface{}) {
 		glog.Infof("Set log verbose level to: %d", level)
 		flag.Set("v", level.String())
 	}
+	newDisableDrain := *newCfg.Spec.DisableDrain
```
zshi-redhat, line 351: This will fail as the initial value for disableDrain (if omitted) is nil.
We could either use a boolean value(not pointer) for DisableDrain (since we don't have to check the existence) or check whether it's null.

roytman: Correct, I used the pointer as an anology with other boolean values in `SriovOperatorConfigSpec`, fixed

zshi-redhat:
looks good to me.
Tested with disableDrain=true, nodes didn't get cordoned.

/cc @pliurh

PR #391, opened 2020-11-02
Bug 1892167: Use node specific data for device plugin config

zshi-redhat:


zshi-redhat:
/cc @atyronesmith 

I'll post the sriov device plugin change tomorrow

atyronesmith (commented):
Looks good.

atyronesmith (approved):

controllers/sriovnetworknodepolicy_controller.go
```diff
@@ -189,15 +190,6 @@ func (r *SriovNetworkNodePolicyReconciler) syncDevicePluginConfigMap(pl *sriovne
 		}
 	} else {
 		logger.Info("ConfigMap already exists, updating")
-		currentConfig := dptypes.ResourceConfList{}
-		err = json.Unmarshal([]byte(string(found.Data[DP_CONFIG_FILENAME])), &currentConfig)
-		if err != nil {
-			return err
-		}
-		if reflect.DeepEqual(currentConfig, data) {
```
pliurh, line 197: Can we sort the map in a predictable order, so that we can use the DeepEqual to check whether it has been updated or not?

zshi-redhat: If it's not updated, r.Update will not update the api data. I think we don't need to check the content explicitly.

PR #393, opened 2020-11-03
WIP: Config sriov switchdev and ovs offloading with MCO

pliurh:
Create a new MachineConfig and MachineConfigPool to insert new systemd
 service swichdev-configuration to selected nodes.

 The swichdev-configuration service creates VFs and configures NIC to
 switchdev mode.


controllers/sriovnetworknodepolicy_controller.go
```diff
@@ -136,11 +149,122 @@ func (r *SriovNetworkNodePolicyReconciler) Reconcile(req ctrl.Request) (ctrl.Res
 		return reconcile.Result{}, err
 	}
 
+	if err = r.syncOffloadMachineConfig(policyList); err != nil {
```
adrianchiris, line 152: this will break deployment in vanilla k8s

pliurh: I'll introduce an environemnt variable to differentiate openshift and vanilla k8s.

bindata/manifests/machine-config/files/configure-switchdev.sh.yaml
```diff
@@ -0,0 +1,51 @@
...
+    sleep 5
+    i=0
+    while read pci_addr num_vfs
+    do
+      # unload VF driver
+      VfDirs=$(ls /sys/bus/pci/devices/${pci_addr} | grep virtfn)
+      for VfDir in $VfDirs
+      do
+        VfPciAddr=$(basename "$( readlink -f /sys/bus/pci/devices/${pci_addr}/$VfDir )")
+        echo $VfPciAddr > /sys/bus/pci/drivers/mlx5_core/unbind || true
```
zshi-redhat, line 30: Can we use a var for driver name instead of hard coded?
for example, the driver info can be written in the `/etc/switchdev.conf` file by policy controller.

bindata/manifests/machine-config/units/ovs-configuration.service.yaml
```diff
@@ -0,0 +1,9 @@
+name: ovs-configuration.service
+dropins:
+- name: 10-switchdev.conf
+  contents: |
+    [Service]
+    ExecStartPre=-/sbin/ip route del default via 10.19.111.254 dev eno2
```
zshi-redhat, line 6: Why this line is needed? It looks like a env specific config.

controllers/sriovnetworknodepolicy_controller.go
```diff
@@ -136,11 +149,122 @@ func (r *SriovNetworkNodePolicyReconciler) Reconcile(req ctrl.Request) (ctrl.Res
...
+	var err error
+	devList := []render.DeviceInfo{}
+	offload := false
+
+	for _, p := range pl.Items {
+		if p.Spec.EswitchMode != sriovnetworkv1.ESWITCHMODE_SWITCHDEV {
+			continue
+		}
+
+		for _, dev := range p.Spec.NicSelector.RootDevices {
```
zshi-redhat, line 172: Since the default PF name is kept as it is after setting to switchdev mode. should this also include pfNames?

controllers/sriovnetworknodepolicy_controller.go
```diff
@@ -136,11 +149,122 @@ func (r *SriovNetworkNodePolicyReconciler) Reconcile(req ctrl.Request) (ctrl.Res
...
+			devList = append(devList, render.DeviceInfo{PciAddress: dev, NumVfs: p.Spec.NumVfs})
```
zshi-redhat, line 173: We shall add the device in devList only when it's on the default ovn network.

controllers/sriovnetworknodepolicy_controller.go
zshi-redhat, line 152: Should we avoid calling this in a non-ovn environment?

controllers/sriovnetworknodepolicy_controller.go
```diff
@@ -136,11 +149,122 @@ func (r *SriovNetworkNodePolicyReconciler) Reconcile(req ctrl.Request) (ctrl.Res
...
+		}
+		offload = true
```
zshi-redhat, line 175: We might want to label the node with `node-role.kubernetes.io/offload-worker` automatically when eswitchdev device is configured on default ovn-network.

bindata/manifests/machine-config/units/ovs-configuration.service.yaml
```diff
@@ -0,0 +1,9 @@
+name: ovs-configuration.service
```
zshi-redhat, line 1: Why this file is needed?

bindata/manifests/machine-config/files/configure-switchdev.sh.yaml
```diff
@@ -0,0 +1,51 @@
...
+    names=()
+
+    while read pci_addr num_vfs
+    do
+      echo "Set $num_vfs VFs on device $pci_addr"
+
+      names+=($(ls /sys/bus/pci/devices/${pci_addr}/net/))
+      # create VFs
+      echo $num_vfs > /sys/bus/pci/devices/${pci_addr}/sriov_numvfs
+    done <"$input"
```
zshi-redhat, line 18: It looks like `input` file contains cluster-wide switchdev device info. should we apply the config according to node specific config ?

controllers/sriovnetworknodepolicy_controller.go
```diff
@@ -61,8 +62,20 @@ func (r *SriovNetworkNodePolicyReconciler) Reconcile(req ctrl.Request) (ctrl.Res
...
+	policy := &sriovnetworkv1.SriovNetworkNodePolicy{}
+	err := r.Get(context.TODO(), req.NamespacedName, policy)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			reqLogger.Info("SriovNetworkNodePolicy resource not found. Ignoring since object must be deleted")
+			return ctrl.Result{}, nil
+		}
+		// Error reading the object - requeue the request.
+		reqLogger.Error(err, "Failed to get SriovNetworkNodePolicy")
+		return ctrl.Result{}, err
```
zshi-redhat, line 74: If the policy is deleted, shall we return nil instead of trying indefinitely?

pkg/utils/utils.go
```diff
@@ -115,6 +118,11 @@ func SyncNodeState(newState *sriovnetworkv1.SriovNetworkNodeState) error {
 		for _, iface := range newState.Spec.Interfaces {
 			if iface.PciAddress == ifaceStatus.PciAddress {
 				configured = true
+				if iface.EswitchMode == sriovnetworkv1.ESWITCHMODE_SWITCHDEV {
+					// Skip for switchdev devices
+					glog.V(2).Infof("SyncNodeState(): skip for switchdev devices")
+					break
```
zshi-redhat, line 124: note: we need to handle syncStatus properly.

bindata/manifests/machine-config/units/ovs-configuration.service.yaml
pliurh: it's for my test environment. will remove it.

controllers/sriovnetworknodepolicy_controller.go
pliurh: Maybe we can do it in the operator webhook.

bindata/manifests/machine-config/units/ovs-configuration.service.yaml
pliurh: it's for my test environment. will remove it.

bindata/manifests/machine-config/files/configure-switchdev.sh.yaml
pliurh: The assumption is we have an identical hardware setup for all the offload workers. All the nodes in one MachineConfigPool will share the identical MachineConfig.

controllers/sriovnetworknodepolicy_controller.go
pliurh: We return nil at line 70.

controllers/sriovnetworknodepolicy_controller.go
pliurh: Yeah, it would be also useful to control the timing of reboot.

bindata/manifests/machine-config/files/switchedev.conf.yaml
```diff
@@ -0,0 +1,6 @@
+mode: 0755
```
zshi-redhat, line 1: why this file cannot be written by sriov config daemon? is it due to the file mode 0755? (we have a node state checkpoint file written by this daemon)
nit: there is an extra `e` in the file name.

pliurh:
replaced by https://github.com/openshift/sriov-network-operator/pull/418

Issue #395, opened 2020-11-03
Suggestion request - remove the need of skopeo in case the env vars present (env.sh)


oshoval:
https://github.com/openshift/sriov-network-operator/blob/release-4.6/hack/env.sh#L1

Looking on env.sh we see skopeo is mandatory since release 4.6,
one can run the cluster without skopeo installed, but with the variables exported
and it would failed, even we actually dont need to use skopeo and to reset the vars.
This might be the situation we will have, in the machines that wont have sudo to install skopeo
(unless its a must for other things).

Option 1:

```
CNI_IMAGE_DIGEST=$(skopeo inspect docker://quay.io/openshift/origin-sriov-cni | jq --raw-output '.Digest')
export SRIOV_CNI_IMAGE=${SRIOV_CNI_IMAGE:-quay.io/openshift/origin-sriov-cni@${CNI_IMAGE_DIGEST}}
```

can be changed to something like
```
if [ -z $CNI_IMAGE_DIGEST ]; then
  CNI_IMAGE_DIGEST=$(skopeo inspect docker://quay.io/openshift/origin-sriov-cni | jq --raw-output '.Digest') 
  export SRIOV_CNI_IMAGE=quay.io/openshift/origin-sriov-cni@${CNI_IMAGE_DIGEST}
fi
```

if it would be done for all the variables, the skopeo need would be removed.
Saving a small time downloading / installing it, in case running in a volatile container,
and not fail in case installing isn't possible.

Option 2: Supply a flag that will override all the need of those skopeo use,
verifying just that all values are set.

wdyt please?
see please ref PR which implements option 2

Thanks

PR #396, opened 2020-11-03
Add an option to skip variable setting in hack/env.sh

oshoval:
In order to allow running without skopeo install,
add support to set SKIP_VAR_SET,
which will skip env settings, skopeo dependency,
and just validate all env vars are set.

Fixes: https://github.com/openshift/sriov-network-operator/issues/395

### DIFF
--- a/hack/env.sh
+++ b/hack/env.sh
@@ -1,22 +1,33 @@
-if ! skopeo -v &> /dev/null
-then
-        echo "skopeo could not be found"
-        exit 1
+if [ -z $SKIP_VAR_SET ]; then
+        if ! skopeo -v &> /dev/null
+        then
+                echo "skopeo could not be found"
+                exit 1
+        fi
+        CNI_IMAGE_DIGEST=$(skopeo inspect docker://quay.io/openshift/origin-sriov-cni | jq --raw-output '.Digest')
+        export SRIOV_CNI_IMAGE=${SRIOV_CNI_IMAGE:-quay.io/openshift/origin-sriov-cni@${CNI_IMAGE_DIGEST}}
+        INFINIBAND_CNI_IMAGE_DIGEST=$(skopeo inspect docker://quay.io/openshift/origin-sriov-infiniband-cni | jq --raw-output '.Digest')
+        export SRIOV_INFINIBAND_CNI_IMAGE=${SRIOV_INFINIBAND_CNI_IMAGE:-quay.io/openshift/origin-sriov-infiniband-cni@${INFINIBAND_CNI_IMAGE_DIGEST}}
+        DP_IMAGE_DIGEST=$(skopeo inspect docker://quay.io/openshift/origin-sriov-network-device-plugin | jq --raw-output '.Digest')
+        export SRIOV_DEVICE_PLUGIN_IMAGE=${SRIOV_DEVICE_PLUGIN_IMAGE:-quay.io/openshift/origin-sriov-network-device-plugin@${DP_IMAGE_DIGEST}}
+        INJECTOR_IMAGE_DIGEST=$(skopeo inspect docker://quay.io/openshift/origin-sriov-dp-admission-controller | jq --raw-output '.Digest')
+        export NETWORK_RESOURCES_INJECTOR_IMAGE=${NETWORK_RESOURCES_INJECTOR_IMAGE:-quay.io/openshift/origin-sriov-dp-admission-controller@${INJECTOR_IMAGE_DIGEST}}
+        DAEMON_IMAGE_DIGEST=$(skopeo inspect docker://quay.io/openshift/origin-sriov-network-config-daemon | jq --raw-output '.Digest')
+        export SRIOV_NETWORK_CONFIG_DAEMON_IMAGE=${SRIOV_NETWORK_CONFIG_DAEMON_IMAGE:-quay.io/openshift/origin-sriov-network-config-daemon@${DAEMON_IMAGE_DIGEST}}
+        WEBHOOK_IMAGE_DIGEST=$(skopeo inspect docker://quay.io/openshift/origin-sriov-network-webhook | jq --raw-output '.Digest')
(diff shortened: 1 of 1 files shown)


hack/env.sh
```diff
@@ -1,22 +1,33 @@
...
+        export NETWORK_RESOURCES_INJECTOR_IMAGE=${NETWORK_RESOURCES_INJECTOR_IMAGE:-quay.io/openshift/origin-sriov-dp-admission-controller@${INJECTOR_IMAGE_DIGEST}}
+        DAEMON_IMAGE_DIGEST=$(skopeo inspect docker://quay.io/openshift/origin-sriov-network-config-daemon | jq --raw-output '.Digest')
+        export SRIOV_NETWORK_CONFIG_DAEMON_IMAGE=${SRIOV_NETWORK_CONFIG_DAEMON_IMAGE:-quay.io/openshift/origin-sriov-network-config-daemon@${DAEMON_IMAGE_DIGEST}}
+        WEBHOOK_IMAGE_DIGEST=$(skopeo inspect docker://quay.io/openshift/origin-sriov-network-webhook | jq --raw-output '.Digest')
+        export SRIOV_NETWORK_WEBHOOK_IMAGE=${SRIOV_NETWORK_WEBHOOK_IMAGE:-quay.io/openshift/origin-sriov-network-webhook@${WEBHOOK_IMAGE_DIGEST}}
+        OPERATOR_IMAGE_DIGEST=$(skopeo inspect docker://quay.io/openshift/origin-sriov-network-operator | jq --raw-output '.Digest')
+        export SRIOV_NETWORK_OPERATOR_IMAGE=${SRIOV_NETWORK_OPERATOR_IMAGE:-quay.io/openshift/origin-sriov-network-operator@${OPERATOR_IMAGE_DIGEST}}
+else
+        [ -z $CNI_IMAGE_DIGEST ] && echo "CNI_IMAGE_DIGEST is empty but SKIP_VAR_SET is set" && exit 1
```
zshi-redhat, line 22: Would this fail when both `SKIP_VAR_SET` && full image url (e.g. SRIOV_CNI_IMAGE) are set?
for example, if user wants to deploy a customized image or self-built image.

oshoval: Nope it wont, it would fail only if SKIP_VAR_SET is set but SRIOV_CNI_IMAGE isnt
lets see this example which pass this line but fail on line 23 as the next VAR isnt set
```
$ export SKIP_VAR_SET=1
$ export CNI_IMAGE_DIGEST=1
INFINIBAND_CNI_IMAGE_DIGEST is empty but SKIP_VAR_SET is set
```

SKIP_VAR_SET skips all the automatic setting, but enforce that each and every env var of the skopeo ones is set in this block

hack/env.sh
```diff
@@ -1,22 +1,33 @@
...
-INJECTOR_IMAGE_DIGEST=$(skopeo inspect docker://quay.io/openshift/origin-sriov-dp-admission-controller | jq --raw-output '.Digest')
-export NETWORK_RESOURCES_INJECTOR_IMAGE=${NETWORK_RESOURCES_INJECTOR_IMAGE:-quay.io/openshift/origin-sriov-dp-admission-controller@${INJECTOR_IMAGE_DIGEST}}
-DAEMON_IMAGE_DIGEST=$(skopeo inspect docker://quay.io/openshift/origin-sriov-network-config-daemon | jq --raw-output '.Digest')
-export SRIOV_NETWORK_CONFIG_DAEMON_IMAGE=${SRIOV_NETWORK_CONFIG_DAEMON_IMAGE:-quay.io/openshift/origin-sriov-network-config-daemon@${DAEMON_IMAGE_DIGEST}}
-WEBHOOK_IMAGE_DIGEST=$(skopeo inspect docker://quay.io/openshift/origin-sriov-network-webhook | jq --raw-output '.Digest')
-export SRIOV_NETWORK_WEBHOOK_IMAGE=${SRIOV_NETWORK_WEBHOOK_IMAGE:-quay.io/openshift/origin-sriov-network-webhook@${WEBHOOK_IMAGE_DIGEST}}
-OPERATOR_IMAGE_DIGEST=$(skopeo inspect docker://quay.io/openshift/origin-sriov-network-operator | jq --raw-output '.Digest')
-export SRIOV_NETWORK_OPERATOR_IMAGE=${SRIOV_NETWORK_OPERATOR_IMAGE:-quay.io/openshift/origin-sriov-network-operator@${OPERATOR_IMAGE_DIGEST}}
+
 export RELEASE_VERSION=4.7.0
```
oshoval, line 31: Question
does
```
CNI_IMAGE_DIGEST=$(skopeo inspect docker://quay.io/openshift/origin-sriov-cni | jq --raw-output '.Digest')
```
takes the latest ? since we dont see in this link the release mentioned.

If so,  should it be something like 
```
export SRIOV_CNI_IMAGE=quay.io/openshift/origin-sriov-cni:${RELEASE_VERSION}
```
in every release, but not on the master?

pliurh: First, openshift users shall always install the operator from the operator hub. And the OKD images for previous releases are used to be out-of-date in the past. It means event you get the latest 4.6 okd images, they are not compiled from the code of the latest 4.6 branch. I am not sure whether it has been changed.  If it does, I think we can add support to install non-latest releases.

hack/env.sh
```diff
@@ -1,22 +1,33 @@
-if ! skopeo -v &> /dev/null
-then
-        echo "skopeo could not be found"
-        exit 1
+if [ -z $SKIP_VAR_SET ]; then
```
oshoval, line 1: is the name fine or should it be named SRIOV_SKIP_VAR_SET ?

pliurh: It's fine.

oshoval:
Thanks @pliurh 
would it be able to backport it please to 4.6 north ?
we would need to bump soon to 4.6+, but not to latest since latest is still developing.

pliurh:
A BZ is required for any backport. To have it merged, we also need approval from the z stream patch manager.

oshoval:
> A BZ is required for any backport. To have it merged, we also need approval from the z stream patch manager.

Ok, thank you

Issue #399, opened 2020-11-06
Add support for incorporating metadata from an OpenStack-based virtual deployment


atyronesmith:
When OpenShift is deployed on OpenStack, OpenStack provides additional [metadata](https://docs.openstack.org/nova/latest/user/metadata.html) about networks that can be used by the sriov-network-operator.  This metadata specifies which OpenStack networks (UUID) are connected to which interface of the VM.  In addition, if [Device Role Tagging](https://specs.openstack.org/openstack/nova-specs/specs/mitaka/approved/virt-device-role-tagging.html) is used, the metadata can also provide labels that allow the grouping of VF resources.

This RFE is contingent upon the approval of [this PR](https://github.com/openshift/sriov-network-operator/pull/369).

An example of the metadata is below:

```json
{
  "uuid": "bc90bd5c-96b8-42a2-a502-6b0b6dbb66af",
  "admin_pass": "...",
  "hostname": "worker-0.fdp.nfv",
  "name": "worker-0.fdp.nfv",
  "launch_index": 0,
  "availability_zone": "nova",
  "random_seed": "....",
  "project_id": "ecdb8ecc548146febe3e1e0145bb6dfb",
  "devices": [
    {
      "vf_trusted": false,
      "type": "nic",
      "mac": "fa:16:3e:5e:d2:74",
      "bus": "pci",
      "address": "0000:00:03.0",
      "tags": [
        "sdn"
      ]
    },
    {
      "vlan": 108,
      "vf_trusted": true,
      "type": "nic",
      "mac": "fa:16:3e:e9:bb:7c",
      "bus": "pci",
      "address": "0000:00:07.0",
      "tags": [
        "uplink"
      ]
    },
    {
      "vlan": 108,
      "vf_trusted": true,
      "type": "nic",
      "mac": "fa:16:3e:8b:7a:97",
      "bus": "pci",
      "address": "0000:00:08.0",
      "tags": [
        "uplink"
      ]
    },
    {
      "vlan": 118,
      "vf_trusted": true,
      "type": "nic",
      "mac": "fa:16:3e:37:a2:44",
      "bus": "pci",
      "address": "0000:00:09.0",
      "tags": [
        "downlink"
      ]
    },
    {
      "vlan": 118,
      "vf_trusted": true,
      "type": "nic",
      "mac": "fa:16:3e:33:c8:51",
      "bus": "pci",
      "address": "0000:00:0a.0",
      "tags": [
        "downlink"
      ]
    },
[...]

PR #400, opened 2020-11-07
Add support for OpenStack metadata

atyronesmith:
[Addresses Issue](https://github.com/openshift/sriov-network-operator/issues/399)

When OpenShift is deployed on OpenStack, OpenStack provides additional metadata about networks that can be used by the sriov-network-operator. This metadata specifies which OpenStack networks (UUID) are connected to which interface of the VM. In addition, if Device Role Tagging is used, the metadata can also provide labels that allow the grouping of VF resources.

### DIFF
--- a/api/v1/helper.go
+++ b/api/v1/helper.go
@@ -3,7 +3,9 @@ package v1
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"os"
+	"regexp"
 	"sort"
 	"strconv"
 	"strings"
@@ -42,6 +44,23 @@ var NicIdMap = []string{
 	"15b3 1017 1018", // ConnectX-5, PCIe 3.0
 	"15b3 101b 101c", // ConnectX-6
 	"15b3 1013 1014", // ConnectX-4
+}
+
+// NetFilterType Represents the NetFilter tags to be used
+type NetFilterType int
+
+const (
+	// OpenstackNetworkID network UUID
+	OpenstackNetworkID NetFilterType = iota
+)
+
+func (e NetFilterType) String() string {
+	switch e {
+	case OpenstackNetworkID:
+		return "openstack/NetworkID"
+	default:
+		return fmt.Sprintf("%d", int(e))
+	}
 }
 
 func IsSupportedVendor(vendorId string) bool {
@@ -274,7 +293,8 @@ func UniqueAppend(inSlice []string, strings ...string) []string {
 // Apply policy to SriovNetworkNodeState CR
 func (p *SriovNetworkNodePolicy) Apply(state *SriovNetworkNodeState, merge bool) {
 	s := p.Spec.NicSelector
-	if s.Vendor == "" && s.DeviceID == "" && len(s.RootDevices) == 0 && len(s.PfNames) == 0 {
+	if s.Vendor == "" && s.DeviceID == "" && len(s.RootDevices) == 0 && len(s.PfNames) == 0 &&
+		len(s.NetFilter) == 0 {
 		// Empty NicSelector match none
 		return
 	}
@@ -430,6 +450,10 @@ func (selector *SriovNetworkNicSelector) Selected(iface *InterfaceExt) bool {
 			return false
 		}
 	}
+	if selector.NetFilter != "" && NetFilterMatch(selector.NetFilter, iface.NetFilter) == false {
+		return false
+	}
+
 	return true
 }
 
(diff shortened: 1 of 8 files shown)


atyronesmith:
Accidental closure....

zshi-redhat:
@atyronesmith  talking to Peng, it might be good to have a generic name for the new selector that can be used across multiple platforms (e.g. vmware, baremetal if needed). do you know if there are equivalent feature in vmware or any other cloud providers for device tagging, device IDs metadata? what are the names used in those platforms?

atyronesmith:
> @atyronesmith talking to Peng, it might be good to have a generic name for the new selector that can be used across multiple platforms (e.g. vmware, baremetal if needed). do you know if there are equivalent feature in vmware or any other cloud providers for device tagging, device IDs metadata? what are the names used in those platforms?

Good point.  Let me check.

zshi-redhat:
> > @atyronesmith talking to Peng, it might be good to have a generic name for the new selector that can be used across multiple platforms (e.g. vmware, baremetal if needed). do you know if there are equivalent feature in vmware or any other cloud providers for device tagging, device IDs metadata? what are the names used in those platforms?
> 
> Good point. Let me check.

In case we don't know what other platforms may support, I think having either one of `networkID` or `networkTag` would be sufficient for openstack VM case, @atyronesmith wdyt? In this PR, we can leave one and remove the other the minimize the API change.

atyronesmith:
> > > @atyronesmith talking to Peng, it might be good to have a generic name for the new selector that can be used across multiple platforms (e.g. vmware, baremetal if needed). do you know if there are equivalent feature in vmware or any other cloud providers for device tagging, device IDs metadata? what are the names used in those platforms?
> > 
> > 
> > Good point. Let me check.
> 
> In case we don't know what other platforms may support, I think having either one of `networkID` or `networkTag` would be sufficient for openstack VM case, @atyronesmith wdyt? In this PR, we can leave one and remove the other the minimize the API change.

How about just infraNet  and have it either be a UUID or a tag?

zshi-redhat:
> > > > @atyronesmith talking to Peng, it might be good to have a generic name for the new selector that can be used across multiple platforms (e.g. vmware, baremetal if needed). do you know if there are equivalent feature in vmware or any other cloud providers for device tagging, device IDs metadata? what are the names used in those platforms?
> > > 
> > > 
> > > Good point. Let me check.
> > 
> > 
> > In case we don't know what other platforms may support, I think having either one of `networkID` or `networkTag` would be sufficient for openstack VM case, @atyronesmith wdyt? In this PR, we can leave one and remove the other the minimize the API change.
> 
> How about just infraNet and have it either be a UUID or a tag?

Does the `Infra` here refer to platform infra such as openstack, vmware etc?

atyronesmith:
> > > > > @atyronesmith talking to Peng, it might be good to have a generic name for the new selector that can be used across multiple platforms (e.g. vmware, baremetal if needed). do you know if there are equivalent feature in vmware or any other cloud providers for device tagging, device IDs metadata? what are the names used in those platforms?
> > > > 
> > > > 
> > > > Good point. Let me check.
> > > 
> > > 
> > > In case we don't know what other platforms may support, I think having either one of `networkID` or `networkTag` would be sufficient for openstack VM case, @atyronesmith wdyt? In this PR, we can leave one and remove the other the minimize the API change.
> > 
> > 
> > How about just infraNet and have it either be a UUID or a tag?
> 
> Does the `Infra` here refer to platform infra such as openstack, vmware etc?

Yes.  it could be called netFilter as well...

zshi-redhat:
> > > > > > @atyronesmith talking to Peng, it might be good to have a generic name for the new selector that can be used across multiple platforms (e.g. vmware, baremetal if needed). do you know if there are equivalent feature in vmware or any other cloud providers for device tagging, device IDs metadata? what are the names used in those platforms?
> > > > > 
> > > > > 
> > > > > Good point. Let me check.
> > > > 
> > > > 
> > > > In case we don't know what other platforms may support, I think having either one of `networkID` or `networkTag` would be sufficient for openstack VM case, @atyronesmith wdyt? In this PR, we can leave one and remove the other the minimize the API change.
> > > 
> > > 
> > > How about just infraNet and have it either be a UUID or a tag?
> > 
> > 
> > Does the `Infra` here refer to platform infra such as openstack, vmware etc?
> 
> Yes. it could be called netFilter as well...

netFilter sounds better than infraNet in my opinion.
@pliurh wdyt

pliurh:
Additionally, we can use a formated string `<platform>/<metadata>:<id>` to make this flag more user friendly, e.g. `openstack/NetworkID:xxxxx-xxxx-xxxx'.

atyronesmith:
> Additionally, we can use a formated string `<platform>/<metadata>:<id>` to make this flag more user-friendly, e.g. `openstack/NetworkID:xxxxx-xxxx-xxxx`.

Yes.  That was my original thought.  However, in general, kubernetes API "doctrine" seems to frown upon "non-well defined" fields.  If you think the above is acceptable, then a more general field would be good.

pliurh:
> > Additionally, we can use a formated string `<platform>/<metadata>:<id>` to make this flag more user-friendly, e.g. `openstack/NetworkID:xxxxx-xxxx-xxxx`.
> 
> Yes. That was my original thought. However, in general, kubernetes API "doctrine" seems to frown upon "non-well defined" fields. If you think the above is acceptable, then a more general field would be good.

Another option is to use `runtime.RawExtension` instead of `string` for this field. This is a way to embed other APIs in a CRD.

zshi-redhat:
> > > Additionally, we can use a formated string `<platform>/<metadata>:<id>` to make this flag more user-friendly, e.g. `openstack/NetworkID:xxxxx-xxxx-xxxx`.
> > 
> > 
> > Yes. That was my original thought. However, in general, kubernetes API "doctrine" seems to frown upon "non-well defined" fields. If you think the above is acceptable, then a more general field would be good.
> 
> Another option is to use `runtime.RawExtension` instead of `string` for this field. This is a way to embed other APIs in a CRD.

@pliurh @atyronesmith 
Should this be discussed later when we have a clear understanding of how other cloud providers are passing network metadata info?  I felt we can start from a simple string value that could already satisfy the need of openstack.

atyronesmith:
I have pushed a commit that contains only a single field, NetFilter.  The netFilter has the format described above: openstack/NetworkID:UUID.  The nodestate shows the netFilter in the format and the nodepolicy can be specified with the format.  
/cc @zshi-redhat @pliurh

atyronesmith:
The PM has agreed that the additional Tag information would cause more confusion than additional value and therefore has been removed.

pliurh:
looks good to me. @zshi-redhat PTAL

PR #402, opened 2020-11-09
add support for Intel NIC X710 10G SFP+

zshi-redhat:


pliurh:
The description in the CRD shall also be updated.

zshi-redhat:
> The description in the CRD shall also be updated.

Done

pliurh:
/lgtm

Note: we need to update the document accordingly.

PR #403, opened 2020-11-10
Use the right object when webhook delete operation

zshi-redhat:
Currently operator-webhook denies every delete operation with:
denied the request: unexpected end of JSON input


zshi-redhat:
re-add the webhook delete operation support for sriovnetworknodepolicy.

PR #408, opened 2020-11-13
Don't update the daemonsets / the network status if there are no changes

fedepaol:
The networkpolicy controller is configured so that daemonsets / sriov network status are subresources of the policy. What happens is that whenever a policy is added, the controller goes into an endless loop where it updates the subresources and get notified of those changes through the primary one (the policy), and then updates again and so on.

This can be seen looking at the operator's logs (I noticed it by chance running the operator locally).

If the subresources' specs do not change from what's read from the api we don't update them, stopping the loop.

### DIFF
--- a/controllers/sriovnetworknodepolicy_controller.go
+++ b/controllers/sriovnetworknodepolicy_controller.go
@@ -28,6 +28,7 @@ import (
 	errs "github.com/pkg/errors"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	uns "k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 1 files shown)


fedepaol:
@zshi-redhat @pliurh PTAL, I think the operator will benefit from this.

zshi-redhat:
@fedepaol thanks for the fix!
So the reason why nodeState is updated continuously is because nodeState.Status is kept being updated.
And for daemonSet, the reason is default field that's not rendered in operator manifests is filled by k8s controllers automatically.

looks good to me.

fedepaol:
> @fedepaol thanks for the fix!
> So the reason why nodeState is updated continuously is because nodeState.Status is kept being updated.
> And for daemonSet, the reason is default field that's not rendered in operator manifests is filled by k8s controllers automatically.
> 

Yes, I noticed it by chance, while I was testing my changes and it wasn't stopping logging

controllers/sriovnetworknodepolicy_controller.go
```diff
@@ -287,13 +290,17 @@ func (r *SriovNetworkNodePolicyReconciler) syncSriovNetworkNodeState(np *sriovne
...
-				p.Apply(found, bool(ppp == p.Spec.Priority))
+				p.Apply(newVersion, bool(ppp == p.Spec.Priority))
 				// record the evaluated policy priority for next loop
 				ppp = p.Spec.Priority
 			}
 		}
-		found.Spec.DpConfigVersion = cksum
-		err = r.Update(context.TODO(), found)
+		newVersion.Spec.DpConfigVersion = cksum
+		if reflect.DeepEqual(found.Spec, newVersion.Spec) {
```
pliurh, line 299: Shall we also user `equality.Semantic.DeepDerivative` here?

fedepaol: I added it in the daemonset as `deepequal` wasn't working, but I guess it won't hurt / will make the code future proof.
Will change

fedepaol: Done

PR #409, opened 2020-11-16
Enable numvfs=0 for VM case

zshi-redhat:


pliurh:
My 2 cents:
1. This change will also allow users to config numvf=0 for baremetal nodes. Users may misconfigure.
2. If we want to allow users to do it on baremetal, we also need to change the behavior on config daemon side, so that the plugin can handle the nodestate.spec properly.

zshi-redhat:
> My 2 cents:
> 
>     1. This change will also allow users to config numvf=0 for baremetal nodes. Users may misconfigure.
> 
I was thinking to add platform discovery also in webhook to prevent user from configuring it on baremetal, it may come later if needed.

>     2. If we want to allow users to do it on baremetal, we also need to change the behavior on config daemon side, so that the plugin can handle the nodestate.spec properly.

We don't want to enable PF case in SR-IOV operator on baremetal, as of now.
numvfs=0 will be only handled by virtual-plugin in VM case, not other plugins.

atyronesmith:
Unless I missed some code, driver selection seems to be VF-based.  While the nodestate shows the driver for the PF (VF==0), the driver change logic is for VFs.  With VF==0, the virtual case would need to modify the driver at what is technically the PF.  It may not really matter, but it seems to be VF functionality that we are emulating, not PF.

zshi-redhat:
PR goes stale,  closing

Issue #394, opened 2020-11-03
Periodic restarting observed for sriov-cni and device-plugin pods


zshi-redhat:
Deploy operator with latest master images and apply sriov node policy.
Wait for ten minutes, check the age of sriov cni and device plugin pods, found that both pods get restarted periodically.

zshi-redhat:
fixed by https://github.com/openshift/sriov-network-operator/pull/422

PR #411, opened 2020-11-17
Bug 1898580: fix the sriov daemon selector assignement.

fedepaol:


fedepaol:
@zshi-redhat @pliurh PTAL

zshi-redhat:
> @zshi-redhat @pliurh PTAL

looks good to me.
I think this might be the reason to issue https://github.com/openshift/sriov-network-operator/issues/394 ?

fedepaol:
> > @zshi-redhat @pliurh PTAL
> 
> looks good to me.
> I think this might be the reason to issue #394 ?

Not sure, I guess it depends on the selector for the policy. With one single selector (or just the default policy) I did not see it happening while trying to reproduce the issue.
In any case, I think the combination of this + https://github.com/openshift/sriov-network-operator/pull/408 will make things more stable.

pliurh:
@fedepaol Could you rebase this PR?

fedepaol:
> @fedepaol Could you rebase this PR?

Done

PR #413, opened 2020-11-20
Bug 1898595: Fix the sriov daemon selector assignement

fedepaol:
This is a manual cherry pick of https://github.com/openshift/sriov-network-operator/pull/411

Signed-off-by: Federico Paolinelli <fpaoline@redhat.com>


fedepaol:
@zshi-redhat @pliurh ptal

sdodson:
Putting a temporary hold on things to prioritize items that fix test flakes while we're hitting AWS quota limits. I'll clear this later today.
/hold

PR #415, opened 2020-11-21
Work around udev race condition related to NM_UNMANAGED=1

andreaskaris:
Work around a udev race condition where NM_UNMANAGED=1 
added to a device path after multiple repeated move operations
of a VF from the host namespace to a pod and back.


fedepaol:
/retitle Bug 1896866: Work around udev race condition related to NM_UNMANAGED=1

PR #416, opened 2020-11-22
Allow non supported hardware for testing purposes

andreaskaris:
This patch implements a new field to the SriovOperatorConfig CRD
with name "unsupportedNicIdMap". This setting allows the configuration
of unsupported NIC models, example: [{"X520":"8086 154d 10ed"}

### DIFF
--- a/api/v1/helper.go
+++ b/api/v1/helper.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"os"
+	"sort"
 	"strconv"
 	"strings"
 
@@ -73,6 +74,77 @@ func IsSupportedModel(vendorId, deviceId string) bool {
 	return false
 }
 
+func IsEnabledUnsupportedVendor(vendorId string, unsupportedNicIdMap map[string]string) bool {
+	for _, n := range unsupportedNicIdMap {
+		if IsValidPciString(n) {
+			ids := strings.Split(n, " ")
+			if vendorId == ids[0] {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+func IsEnabledUnsupportedDevice(deviceId string, unsupportedNicIdMap map[string]string) bool {
+	for _, n := range unsupportedNicIdMap {
+		if IsValidPciString(n) {
+			ids := strings.Split(n, " ")
+			if deviceId == ids[1] {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+func IsEnabledUnsupportedModel(vendorId, deviceId string, unsupportedNicIdMap map[string]string) bool {
+	for _, n := range unsupportedNicIdMap {
+		if IsValidPciString(n) {
+			ids := strings.Split(n, " ")
+			if vendorId == ids[0] && deviceId == ids[1] {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+func IsValidPciString(nicIdString string) bool {
+	ids := strings.Split(nicIdString, " ")
+
+	if len(ids) != 3 {
+		log.Info("IsValidPciString(): ", nicIdString)
+		return false
+	}
+
+	if len(ids[0]) != 4 {
+		log.Info("IsValidPciString():", "Invalid vendor PciId ", ids[0])
+		return false
+	}
+	if _, err := strconv.ParseInt(ids[0], 16, 32); err != nil {
+		log.Info("IsValidPciString():", "Invalid vendor PciId ", ids[0])
+	}
+
+	if len(ids[1]) != 4 {
+		log.Info("IsValidPciString():", "Invalid PciId of PF ", ids[1])
+		return false
+	}
+	if _, err := strconv.ParseInt(ids[1], 16, 32); err != nil {
+		log.Info("IsValidPciString():", "Invalid PciId of PF ", ids[1])
+	}
+
+	if len(ids[2]) != 4 {
+		log.Info("IsValidPciString():", "Invalid PciId of VF ", ids[2])
+		return false
+	}
+	if _, err := strconv.ParseInt(ids[2], 16, 32); err != nil {
+		log.Info("IsValidPciString():", "Invalid PciId of VF ", ids[2])
+	}
+
+	return true
+}
+
 func GetSupportedVfIds() []string {
 	var vfIds []string
 	for _, n := range NicIdMap {
@@ -83,6 +155,49 @@ func GetSupportedVfIds() []string {
 		}
 	}
 	return vfIds
+}
+
+func GetUnsupportedVfIds(unsupportedNicIdMap map[string]string) []string {
+	var vfIds []string
+	for k, n := range unsupportedNicIdMap {
+		if !IsValidPciString(n) {
+			log.Info("GetUnsupportedVfIds():", "name", k,
+				"Invalid Pci string", n)
+			continue
+		}
+		ids := strings.Split(n, " ")
+		vfId := "0x" + ids[2]
+		if !StringInArray(vfId, vfIds) {
+			vfIds = append(vfIds, vfId)
+		}
+	}
+	return vfIds
+}
+
+func GetMergedVfIds(unsupportedNicIdMap map[string]string) []string {
+	supportedVfIds := VfIds
+	unsupportedVfIds := GetUnsupportedVfIds(unsupportedNicIdMap)
+	var mergedVfIds []string
+
+	mergedVfIdsSet := make(map[string]struct{})
+	for _, v := range supportedVfIds {
+		mergedVfIdsSet[v] = struct{}{}
+	}
+	for _, v := range unsupportedVfIds {
+		mergedVfIdsSet[v] = struct{}{}
+	}
+	for k := range mergedVfIdsSet {
+		mergedVfIds = append(mergedVfIds, k)
+	}
+
+	// return a sorted slice so that udev rule is stable
+	sort.Slice(mergedVfIds, func(i, j int) bool {
+		ip, _ := strconv.ParseInt(mergedVfIds[i], 0, 32)
+		jp, _ := strconv.ParseInt(mergedVfIds[j], 0, 32)
+		return ip < jp
+	})
+
+	return mergedVfIds
 }
 
 func GetVfDeviceId(deviceId string) string {
--- a/deploy/role.yaml
+++ b/deploy/role.yaml
@@ -81,3 +81,22 @@ rules:
     - securitycontextconstraints
   verbs:
     - use
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
+  verbs:
+  - get
+  - update
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: operator-webhook-sa
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
+  verbs:
+  - get
--- a/deploy/role_binding.yaml
+++ b/deploy/role_binding.yaml
@@ -21,3 +21,15 @@ roleRef:
   kind: Role
   name: sriov-network-config-daemon
   apiGroup: rbac.authorization.k8s.io
+---
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: operator-webhook-sa
+subjects:
+- kind: ServiceAccount
+  name: operator-webhook-sa
+roleRef:
+  kind: Role
+  name: operator-webhook-sa
+  apiGroup: rbac.authorization.k8s.io
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -20,6 +20,7 @@ import (
 	"github.com/golang/glog"
 	"golang.org/x/time/rate"
 	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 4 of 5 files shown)


andreaskaris:
I just realized that it can take up to 30 seconds for the udev rule to be updated (due to the timer firing every 30 seconds).

[root@openshift-worker-0 ~]# cat /etc/udev/rules.d/10-nm-unmanaged.rules 
ACTION=="add|change", ATTRS{device}=="0x1014|0x1016|0x1018|0x101c|0x154c", ENV{NM_UNMANAGED}="1"
[root@openshift-worker-0 ~]# cat /etc/udev/rules.d/10-nm-unmanaged.rules 
ACTION=="add|change", ATTRS{device}=="0x1014|0x1016|0x1018|0x101c|0x10ed|0x10ee|0x10ef|0x154c", ENV{NM_UNMANAGED}="1"

But the admission controller will allow the new device right away. So, there's a 30 seconds race where the udev rule might not yet be in place, but the admission controller webhook will admit the device.

I think that `validateNicModel` should call yet another method that checks the state of the udev rule on disk, as well, and it should make sure that the file was properly updated.

I'll have to amend this one of these days.

zshi-redhat:
@andreaskaris thanks for adding this feature for unsupport NICs.
Several questions:
1) If we were to extend SriovOperatorConfig CRD to add unsupportedNicIdMap field, why need a separate configMap to store the data?
2) How will `X520` in example [{"X520":"8086 154d 10ed"} be used in Operator?

andreaskaris:
@zshi-redhat 


>  If we were to extend SriovOperatorConfig CRD to add unsupportedNicIdMap field, why need a separate configMap to store the data?

Agreed. I had started out modifying the CRD, but then I was not sure how to access it, and then went on to use configmaps. I was not aware that I could directly read the SriovOperatorConfig CRD from the webhook and the network-config-daemon and thought I'd have to generate a configmap first, and then read it from the other daemons. 

What do you think about me simply getting rid of the changes to the CRD, and instead just using a configmap with a pre-agreed name. Admins would have to create a configmap with that exact name, so for example "unsupported-nic-id' (at least that's what I'm using at the moment). That way, it's also a hidden feature, and it introduces less code changes. And it comes with the advantage that I only have to throw away part of this patch, whereas I can keep everything in the webhook and the sriov-network-config-daemon.

> How will X520 in example [{"X520":"8086 154d 10ed"} be used in Operator?

If we are not going with a configmap, but just with a CRD, we could use a simple list instead of a map.  But configmaps must be of type map[string]string:
https://kubernetes.io/docs/concepts/configuration/configmap/
~~~
ConfigMap object
A ConfigMap is an API object that lets you store configuration for other objects to use. Unlike most Kubernetes objects that have a spec, a ConfigMap has data and binaryData fields. These fields accepts key-value pairs as their values. Both the data field and the binaryData are optional. The data field is designed to contain UTF-8 byte sequences while the binaryData field is designed to contain binary data.
~~~
So, users must provide a key, which I think should be a human readable identifier to identify the NIC's name. Hence, "X520" in this example is both a human readable identifier, and a requirement because we need some unique key if we are using a configmap.

andreaskaris:
Hi,

I'm curretly working on an amended version of my patch. I'll update here as soon as that's done, it might tage me a couple of days. For the race condition which I mentioned earlier, I will use timestamp annotations to the configmap. And I will drop the CRD modification and only go with the configmap solution.

- Andreas

andreaskaris:
Hi,

How to use this feature. The following is a configmap that enables the X520 (with correct data) and for the X540 and X560 entries I just added some dummy data ...
~~~
cat <<'EOF' > configmap.yaml
apiVersion: v1
data:
  X520: 8086 154d 10ed
  X540: 8086 154d 10ee
  X560: 8086 154d 10ef
kind: ConfigMap
metadata:
  creationTimestamp: "2020-11-21T23:32:53Z"
  name: unsupported-nic-ids
  namespace: openshift-sriov-network-operator
EOF
~~~

oc apply -f configmap.yaml ; sleep 5 ; oc get configmap unsupported-nic-ids -o yaml ; oc apply -f networkpolicy-netdevice.yaml

For as long as the configmap was not yet annotated correctly, this will fail. In order to avoid race conditions when initially creating the configmap or modifying it, I push the configmap's content into annotations:
~~~
[root@openshift-jumpserver-0 sriov]# oc get configmap unsupported-nic-ids -o yaml
apiVersion: v1
data:
  X520: 8086 154d 10ed
  X540: 8086 154d 10ee
  X560: 8086 154d 10ef
kind: ConfigMap
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"v1","data":{"X520":"8086 154d 10ed","X540":"8086 154d 10ee","X560":"8086 154d 10ef"},"kind":"ConfigMap","metadata":{"annotations":{},"creationTimestamp":"2020-11-21T23:32:53Z","name":"unsupported-nic-ids","namespace":"openshift-sriov-network-operator"}}
    openshift.io/openshift-worker-0: '{"X520":"8086 154d 10ed","X540":"8086 154d 10ee","X560":"8086 154d 10ef"}'
    openshift.io/openshift-worker-1: '{"X520":"8086 154d 10ed","X540":"8086 154d 10ee","X560":"8086 154d 10ef"}'
  creationTimestamp: "2020-11-24T12:23:00Z"
  managedFields:
  - apiVersion: v1
    fieldsType: FieldsV1
    fieldsV1:
      f:data:
        .: {}
        f:X520: {}
        f:X540: {}
        f:X560: {}
      f:metadata:
        f:annotations:
          .: {}
          f:kubectl.kubernetes.io/last-applied-configuration: {}
    manager: kubectl-client-side-apply
    operation: Update
    time: "2020-11-24T12:23:00Z"
  - apiVersion: v1
    fieldsType: FieldsV1
    fieldsV1:
      f:metadata:
        f:annotations:
          f:openshift.io/openshift-worker-0: {}
          f:openshift.io/openshift-worker-1: {}
    manager: sriov-network-config-daemon
    operation: Update
    time: "2020-11-24T12:23:39Z"
  name: unsupported-nic-ids
  namespace: openshift-sriov-network-operator
  resourceVersion: "1387685"
  selfLink: /api/v1/namespaces/openshift-sriov-network-operator/configmaps/unsupported-nic-ids
  uid: 32c3fed1-3fed-4f2c-9eb2-7c1edacdb234
~~~

If the annotation and the configmap's content are equal for a given worker, then the admission webhook knows that the change was already applied to that worker and returns 'true'.

There's one last outstanding issue, but I don't know if this needs to be addressed given that this is an unsupported feature. When deleting the configmap *before* deleting the network policies in place, then the webhook will complain upon deletion. That's because the webhook runs on both *adding* a network policy and on *deleting* a network policy. However, there's no need to run a nic compatibility check on network policy deletion, but that logic was already in place before:
~~~
[root@openshift-jumpserver-0 sriov]# oc delete configmap unsupported-nic-ids
configmap "unsupported-nic-ids" deleted
[root@openshift-jumpserver-0 sriov]# oc delete -f networkpolicy-netdevice.yaml
Error from server (no supported NIC is selected by the nicSelector in CR sriov-enp5s0f1-netdevice): error when deleting "networkpolicy-netdevice.yaml": admission webhook "operator-webhook.sriovnetwork.openshift.io" denied the request: no supported NIC is selected by the nicSelector in CR sriov-enp5s0f1-netdevice
~~~

andreaskaris:
So, IMO func validateSriovNetworkNodePolicy should always return true when operation == DELETE , unless its the default policy.

I would change:
~~~
func validateSriovNetworkNodePolicy(cr *sriovnetworkv1.SriovNetworkNodePolicy, operation v1beta1.Operation) (bool, []string, error) {
        glog.V(2).Infof("validateSriovNetworkNodePolicy: %v", cr)
        var warnings []string

        if cr.GetName() == "default" && cr.GetNamespace() == os.Getenv("NAMESPACE") {
                if operation == "DELETE" {
                        // reject deletion of default policy
                        return false, warnings, fmt.Errorf("default SriovNetworkNodePolicy shouldn't be deleted")
                } else {
                        // skip validating default policy
                        return true, warnings, nil
                }
        }

        if cr.GetNamespace() != os.Getenv("NAMESPACE") {
                warnings = append(warnings, cr.GetName()+
                        " is created or updated but not used. Only policy in openshift-sriov-network-operator namespace is respected.")
        }

        admit, err := staticValidateSriovNetworkNodePolicy(cr)
        if err != nil {
                return admit, warnings, err

        }

        admit, err = dynamicValidateSriovNetworkNodePolicy(cr)
        if err != nil {
                return admit, warnings, err

        }

        return admit, warnings, nil
}
~~~

To:
~~~
func validateSriovNetworkNodePolicy(cr *sriovnetworkv1.SriovNetworkNodePolicy, operation v1beta1.Operation) (bool, []string, error) {
        glog.V(2).Infof("validateSriovNetworkNodePolicy: %v", cr)
        var warnings []string

        if cr.GetName() == "default" && cr.GetNamespace() == os.Getenv("NAMESPACE") {
                if operation == "DELETE" {
                        // reject deletion of default policy
                        return false, warnings, fmt.Errorf("default SriovNetworkNodePolicy shouldn't be deleted")
                } else {
                        // skip validating default policy
                        return true, warnings, nil
                }
        }

        if cr.GetNamespace() != os.Getenv("NAMESPACE") {
                warnings = append(warnings, cr.GetName()+
                        " is created or updated but not used. Only policy in openshift-sriov-network-operator namespace is respected.")
        }

        if operation == "DELETE" {
                return true, warnings, nil
        }

        admit, err := staticValidateSriovNetworkNodePolicy(cr)
        if err != nil {
                return admit, warnings, err

        }

        admit, err = dynamicValidateSriovNetworkNodePolicy(cr)
        if err != nil {
                return admit, warnings, err

        }

        return admit, warnings, nil
}
~~~

andreaskaris:
Also, commands that I used to validate:

cat <<'EOF' > configmap.yaml
apiVersion: v1
data:
  X520: 8086 154d 10ed
  X540: 8086 154d 10ee
  X560: 8086 154d 10ef
kind: ConfigMap
metadata:
  creationTimestamp: "2020-11-21T23:32:53Z"
  name: unsupported-nic-ids
  namespace: openshift-sriov-network-operator
EOF

for p in $(oc get pods -o name | grep network-config-daemon) ; do oc logs --tail=0 -f $p | grep -i period -A10 &  done
for p in $(oc get pods -o name | grep webhook) ; do oc logs --tail=0 -f $p &  done

oc apply -f networkpolicy-netdevice.yaml
oc apply -f configmap.yaml ; sleep 5 ; oc get configmap unsupported-nic-ids -o yaml ; oc apply -f networkpolicy-netdevice.yaml
sleep 30
oc apply -f networkpolicy-netdevice.yaml



cat <<'EOF' > configmap.yaml
apiVersion: v1
data:
  X540: 8086 154d 10ee
kind: ConfigMap
metadata:
  creationTimestamp: "2020-11-21T23:32:53Z"
  name: unsupported-nic-ids
  namespace: openshift-sriov-network-operator
EOF
oc apply -f configmap.yaml ; sleep 1 ; oc get configmap unsupported-nic-ids -o yaml ; oc apply -f networkpolicy-netdevice.yaml

akaris-redhat:
@zshi-redhat Do I need to do anything else to get this in? The tests all passed, but I assume that I need some more code review. Let me know if the code needs to be improved.

Thanks!

zshi-redhat:
> @zshi-redhat Do I need to do anything else to get this in? The tests all passed, but I assume that I need some more code review. Let me know if the code needs to be improved.
> 
> Thanks!

@andreaskaris hi I took a look last week, it looks generally good to me.
The approach it takes to use configmap aligns with what we discussed before in sriov team.

One comment is about the update of configmap annotation. Since each config daemon will need to update single configmap, do you see there might be collision in a large cluster (although this will eventually be sorted out as it updates configmap every 20 seconds)?  Maybe there already are examples from other operator doing similar update?

About writing udev files, we overwrite it every 30 seconds if I read it correct, shall we compare and only write when there is a change?

akaris-redhat:
>> One comment is about the update of configmap annotation. Since each config daemon will need to update single configmap, do you see there might be collision in a large cluster (although this will eventually be sorted out as it updates configmap every 30 seconds)? Maybe there already are examples from other operator doing similar update?

I thought that multiple clients could write to the same configmap simultaneously, i.e. kubernetes would take care of this. Do I need to implement a retry mechanism somehow? Are there any examples in the current code? TBH, given that this  configmap will only exist in unsupported systems, I don't even know if we should worry too much about this?

>> About writing udev files, it is overwritten every 30 seconds if I read it correct, shall we compare and only write when there is a change?

I will implement that.

akaris-redhat:
Ok so now:
* the operator only writes to the udev file if contents change in udev file
* the operator only updates the configmap if contents change in udev file

akaris-redhat:
I also addressed that:

akaris-redhat:
@zshi-redhat Please let me know if there are further suggestions :-)

zshi-redhat:
> @zshi-redhat Please let me know if there are further suggestions :-)

nope, thank you for contributing the feature!

/lgtm

andreaskaris:
Just because the configmap's format was ruined by the copy / paste action earlier, here is the configmap:
~~~
apiVersion: v1
data:
  X520: 8086 154d 10ed
  X540: 8086 154d 10ee
  X560: 8086 154d 10ef
kind: ConfigMap
metadata:
  name: unsupported-nic-ids
  namespace: openshift-sriov-network-operator
~~~

andreaskaris:
The format per entry is:
~~~
[human readable name]: [vendor ID] [PF device ID] [VF devide ID]
~~~

PR #418, opened 2020-11-24
Use MCO to apply switchdev configuration

pliurh:
1. Add flag 'enableOvnOffload' in SriovOperatorConfig, operator will
create MC and MCP for injectin switchdev-configuratio.service when
it's set to 'true'.
2. switchdev-configuratio.service creates VFs and configures PFs to
switchdev mode.
3. Add 'mco-plugin' in the config daemon, which 1) generates 'switchdev.conf'
file consumed by switchdev-configuratio.service; 2) put node in/out
the offload MCP.
4. Use environment variable CLUSTER_TYPE to differentiate cluster types.


api/v1/sriovoperatorconfig_types.go
```diff
@@ -22,6 +22,8 @@ type SriovOperatorConfigSpec struct {
 	LogLevel int `json:"logLevel,omitempty"`
 	// Flag to disable nodes drain during debugging
 	DisableDrain bool `json:"disableDrain,omitempty"`
+	// Flag to enable OVN hardware offload
+	EnableOvnOffload bool `json:"enableOvnOffload,omitempty"`
```
zshi-redhat, line 26: Use OVS instead of OVN? I felt sometimes OVN offload is referring to smart NIC offload.
Also does it make sense to not expose the boolean flag ?
for example, if switchdev poilicy is configured on node interface with default route, then render the MCO manifests.
If switchdev policy is configured on non-default interface (switchdev CNI in the future), then don't render MCO manifests.

controllers/sriovoperatorconfig_controller.go
```diff
@@ -522,3 +531,97 @@ func (r *SriovOperatorConfigReconciler) syncK8sResource(cr *sriovnetworkv1.Sriov
 	}
 	return nil
 }
+
+func (r *SriovOperatorConfigReconciler) syncOffloadMachineConfig(dc *sriovnetworkv1.SriovOperatorConfig) error {
```
zshi-redhat, line 535: Shall we split the file renderring to two parts and sync the offload in sriov policy controller?
for example, one part is switchdev related configuration files (switchdev-configuration service, configure-switchdev script and NetworkManager service update)
the other part is ovs offload related configuration files (MCP, MC, ovs-vswitchd)

This way, if we were to add switchdev CNI support in the future, then it will not associated with the ovs-configuration and MCO changes.

bindata/manifests/machine-config/files/configure-switchdev.sh.yaml
```diff
@@ -0,0 +1,56 @@
...
+    sleep 5
+    i=0
+    while read pci_addr num_vfs
+    do
+      # unload VF driver
+      VfDirs=$(ls /sys/bus/pci/devices/${pci_addr} | grep virtfn)
+      for VfDir in $VfDirs
+      do
+        VfPciAddr=$(basename "$( readlink -f /sys/bus/pci/devices/${pci_addr}/$VfDir )")
+        echo $VfPciAddr > /sys/bus/pci/drivers/mlx5_core/unbind || true
```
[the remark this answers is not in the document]
pliurh: As the switchdev enablement procedure could be different for different devices, I double whether it makes sense to make it a variable.

Dockerfile.rhel7
```diff
@@ -13,4 +13,5 @@ COPY --from=builder /go/src/github.com/openshift/sriov-network-operator/build/_o
 COPY --from=builder /go/src/github.com/openshift/sriov-network-operator/manifests /manifests
 COPY bindata /bindata
 ENV OPERATOR_NAME=sriov-network-operator
+ENV CLUSTER_TYPE=openshift
```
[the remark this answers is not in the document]
pliurh: My consideration is that we can reuse the operator manifests for both upstream and downstream, and we already have different Dockerfile for openshift and vanilla k8s.

hack/env.sh
```diff
@@ -32,3 +32,4 @@ export RELEASE_VERSION=4.7.0
 export OPERATOR_NAME=sriov-network-operator
 export RESOURCE_PREFIX=openshift.io
 export ENABLE_ADMISSION_CONTROLLER=${ENABLE_ADMISSION_CONTROLLER:-"true"}
+export CLUSTER_TYPE=${CLUSTER_TYPE:-openshift}
```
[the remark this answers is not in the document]
pliurh: This will only be used by `make run` which is only for development usage. If developers want to run in vanilla k8s, they can run `CLUSTER_TYPE=kubernetes make run`. In the deploy case, we already have this environment variable in the openshift images.

pkg/utils/utils.go
```diff
@@ -118,6 +123,10 @@ func SyncNodeState(newState *sriovnetworkv1.SriovNetworkNodeState) error {
 		for _, iface := range newState.Spec.Interfaces {
 			if iface.PciAddress == ifaceStatus.PciAddress {
 				configured = true
+				if iface.EswitchMode == sriovnetworkv1.ESWITCHMODE_SWITCHDEV && ClusterType == "openshift" {
```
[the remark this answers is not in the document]
pliurh: Correct.

api/v1/sriovoperatorconfig_types.go
pliurh: OVS offload could imply the openshift-sdn can also be offloaded.

If we use the nodepolicy to control the MC and MCP, it would be difficult to synchronize the reboot logic between MCO and SNO.

controllers/sriovoperatorconfig_controller.go
pliurh: fixed.

bindata/manifests/machine-config/machineconfigpool.yaml
```diff
@@ -0,0 +1,12 @@
+---
+apiVersion: machineconfiguration.openshift.io/v1
+kind: MachineConfigPool
+metadata:
+  name: offload
```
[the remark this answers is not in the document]
pliurh: fixed

bindata/manifests/machine-config/machineconfigpool.yaml
```diff
@@ -0,0 +1,12 @@
...
+spec:
+  machineConfigSelector:
+    matchExpressions:
+      - {key: machineconfiguration.openshift.io/role, operator: In, values: [worker,offload]}
+  nodeSelector:
+    matchLabels:
+      node-role.kubernetes.io/offload-worker: ""
```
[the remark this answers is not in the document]
pliurh: fixed

bindata/manifests/machine-config/units/switchdev-configuration.yaml
```diff
@@ -0,0 +1,19 @@
+contents: |
+  [Unit]
+  Description=Configures SRIOV NIC into swtichdev mode
+  # Removal of this file signals firstboot completion
+  ConditionPathExists=!/etc/ignition-machine-config-encapsulated.json
+  # This service is used to move a SRIOV NIC into swtichdev mode
```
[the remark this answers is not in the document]
pliurh: fixed

pliurh:
cc @adrianchiris

bindata/manifests/machine-config/units/switchdev-configuration.yaml
```diff
@@ -0,0 +1,19 @@
+contents: |
+  [Unit]
+  Description=Configures SRIOV NIC into swtichdev mode
```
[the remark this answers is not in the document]
pliurh: fixed

pkg/plugins/mco/mco_plugin.go
```diff
@@ -0,0 +1,203 @@
...
+			glog.Infof("Move node %s into HW offload MachineConfigPool", node.Name)
+			mergePatch, _ := json.Marshal(map[string]interface{}{
+				"metadata": map[string]interface{}{
+					"labels": map[string]interface{}{
+						label: "",
+					},
+				},
+			})
+			kubeclient.CoreV1().Nodes().Patch(context.Background(), nodeName, types.MergePatchType, mergePatch, metav1.PatchOptions{})
+		}
```
zshi-redhat, line 137: return or add `else` here? otherwise the next glog message will always be printed.

pkg/plugins/mco/mco_plugin.go
```diff
@@ -0,0 +1,203 @@
...
+
+// Apply config change
+func (p *McoPlugin) Apply() error {
+	glog.Info("mco-plugin Apply()")
+	node, err := kubeclient.CoreV1().Nodes().Get(context.Background(), nodeName, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+	label := nodeLabelPrefix + controllers.HwOffloadNodeLabel
+	if switchdevConfigured {
```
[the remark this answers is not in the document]
pliurh: No, they are different. `switchdevConfigured` indicates whether swichdev is configured for any interface. `update` indicates whether the file `switchdev.conf` is changed.

pkg/plugins/mco/mco_plugin.go
```diff
@@ -0,0 +1,203 @@
...
+	glog.Info("mco-plugin OnNodeStateChange()")
+	switchdevConfigured = false
+	for _, iface := range new.Spec.Interfaces {
+		if iface.EswitchMode == sriovnetworkv1.ESWITCHMODE_SWITCHDEV {
+			switchdevConfigured = true
+		}
+	}
+
+	var update, remove bool
+	if update, remove, err = writeSwitchdevConfFile(new); err != nil {
```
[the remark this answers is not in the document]
pliurh: fixed

pkg/plugins/mco/mco_plugin.go
```diff
@@ -0,0 +1,203 @@
...
+		return
+	}
+	if newContent == string(oldContent) {
+		glog.V(2).Info("writeSwitchdevConfFile(): no update")
+		return
+	}
+	if newContent == "" {
+		remove = true
+		glog.V(2).Info("writeSwitchdevConfFile(): remove switchdev.conf")
+		if err = os.Remove(switchDevConfPath); err != nil {
```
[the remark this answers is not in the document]
pliurh: fixed.

pkg/plugins/mco/mco_plugin.go
pliurh: fixed.

api/v1/sriovoperatorconfig_types.go
adrianchiris: So for OVS offload on secondary network we would have it in nodepolicy ? 
(ovn is not expected to be used in this case)

pkg/plugins/mco/mco_plugin.go
```diff
@@ -0,0 +1,200 @@
...
+				glog.Errorf("writeSwitchdevConfFile(): fail to create file: %v", err)
+				return
+			}
+		} else {
+			return
+		}
+	}
+	newContent := ""
+	for _, iface := range newState.Spec.Interfaces {
+		if iface.EswitchMode == sriovnetworkv1.ESWITCHMODE_SWITCHDEV {
```
Mmduh-483, line 175: Add a condition that the interface support changing eswitchMode, the state interface has eswitchMode !=""

controllers/sriovoperatorconfig_controller.go
```diff
@@ -119,6 +122,12 @@ func (r *SriovOperatorConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Result
 		return reconcile.Result{}, err
 	}
 
+	if utils.ClusterType == "openshift" {
```
Mmduh-483, line 125: "openshift" ClusterType is appearing in several locations, consider adding a constant to utils

pkg/plugins/generic/generic_plugin.go
```diff
@@ -186,13 +186,24 @@ func needDrainNode(desired sriovnetworkv1.Interfaces, current sriovnetworkv1.Int
 		for _, iface := range desired {
 			if iface.PciAddress == ifaceStatus.PciAddress {
 				configured = true
-				if iface.NumVfs != ifaceStatus.NumVfs || iface.Mtu != ifaceStatus.Mtu {
+				if iface.EswitchMode == sriovnetworkv1.ESWITCHMODE_SWITCHDEV {
```
Mmduh-483, line 189: Add check for ClusterType is openshift

adrianchiris (commented):
Im thinking we can reuse the service files and switchdev.conf for vanilla k8s as well, only have config daemon to create the service files instead of relying on mco.

@Mmduh-483 will be working on this base on this PR

Im also wondering if we would benefit from moving other operation to system services. as having controll on when to run certain operations would make our lives easier.
e.g on machine boot.

api/v1/sriovoperatorconfig_types.go
zshi-redhat: @adrianchris could you clarify the use case of ovs offload on secondary network? is it using ovs-cni if not ovn?

pkg/plugins/mco/mco_plugin.go
pliurh: Is there a way to check whether an interface support changing eswitchMode or not? EswitchMode in the NodeState can only be either 'switchdev' or ''.

pkg/plugins/generic/generic_plugin.go
pliurh: I think for vanilla k8s, we also want to use a systemd service to configure the switch device. So from the generic plugin perspective, the logic shall be the same.

pkg/plugins/mco/mco_plugin.go
```diff
@@ -0,0 +1,200 @@
...
+}
+
+// OnNodeStateChange Invoked when SriovNetworkNodeState CR is updated, return if need dain and/or reboot node
+func (p *McoPlugin) OnNodeStateChange(old, new *sriovnetworkv1.SriovNetworkNodeState) (needDrain bool, needReboot bool, err error) {
+	glog.Info("mco-plugin OnNodeStateChange()")
+	switchdevConfigured = false
+	for _, iface := range new.Spec.Interfaces {
+		if iface.EswitchMode == sriovnetworkv1.ESWITCHMODE_SWITCHDEV {
+			switchdevConfigured = true
+		}
```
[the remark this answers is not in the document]
pliurh: fixed

controllers/sriovoperatorconfig_controller.go
pliurh: fixed

config/crd/bases/sriovnetwork.openshift.io_sriovoperatorconfigs.yaml
```diff
@@ -53,6 +53,9 @@ spec:
                 description: Flag to control whether the operator admission controller
                   webhook shall be deployed
                 type: boolean
+              enableOvnOffload:
+                description: Flag to enable OVN hardware offload
```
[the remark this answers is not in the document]
pliurh: done

api/v1/sriovoperatorconfig_types.go
pliurh: @adrianchiris As the offload is enabled in OVS, can we say that all the traffic on OVS can be offloaded, no matter what control plane it connects to? If yes, I think we shall change this flag to `EnableOvsOffload`.

adrianchiris: @pliurh yes, i think it should be `EnableOvsOffload` , the SDN is not relevant here IMO.

@zshi-redhat it may be

zshi-redhat: The configuration is for sriov node only. I expect it will control the ovs configuration (hw-offload, tc-policy). so it will be used only for hw-offload use case. For other cases like switchdev-cni that doesn't require ovs bridge, we rely on config daemon to configure the device.

pliurh: changed the flag name to `EnableOvsOffload`.

pkg/plugins/mco/mco_plugin.go
```diff
@@ -0,0 +1,201 @@
+package main
```
Mmduh-483, line 1: The same code will be used for vanilla k8s, can you rename it to switchdev_plugin, code here is only related to switchdev, it doesn't recognize how the service injected

bindata/manifests/machine-config/files/configure-switchdev.sh.yaml
```diff
@@ -0,0 +1,56 @@
...
+
+      done
+
+      # set PF to switchdev mode
+      devlink dev eswitch set pci/${pci_addr} mode switchdev
+
+      # reset the pf name
+      new_name=$(ls /sys/bus/pci/devices/${pci_addr}/net/)
```
adrianchiris, line 43: you should also  enable tc-hw-offload via ethtool

`/usr/sbin/ethtool -K $(new_name) hw-tc-offload on`

pkg/plugins/mco/mco_plugin.go
pliurh: This plugin does 2 tasks: 1) generate switchdev.conf file; 2) move the node to an MCP. The second task is only for MCO. I think for vanilla k8s, we can create a separte plugin to do 1) provision the swichdev.configuratio.service; 2) generate switchdev.conf file. And we can load plugins based on the ClusterType. The logic of generating swichdev.conf can be reused in both plugins. But I think it can be addressed later.

bindata/manifests/machine-config/files/configure-switchdev.sh.yaml
pliurh: In my environment, `hw-tc-offload` is on after swichdev enabled. And ethtool is not available in RHCOS.

zhaozhanqi: @pliurh 
seems ethtool already exist in 4.7 rhcos 
```
which ethtool 
/usr/sbin/ethtool
```

adrianchiris: i see,
on older platform it was required to also enable via ethtool.

PR #419, opened 2020-11-25
Tests: ignore get errors when checking for sriov stable

fedepaol:
When running the tests against a single cluster, configuration changes may trigger reboots. Here we ignore the error, assuming that the cluster will be back within the timeout. Other calls to the api server are done before reaching this stage, so it's safe to assume that running the tests against a wrong cluster will result in an earlier error.


SchSeba (approved):
/lgtm

PR #420, opened 2020-11-25
Automatically copy generated yaml files to the ./manifests/4.7 dir

roytman:
Automatically copies generated `yaml` files from `./config/crd/bases` to `./manifests/4.7`

**Note:**  the  `./manifests/4.7` dirictory conatins 3 additional files, I've not found thier sources, so they probably should be continue to be manually updated.
The files are:

- sriov-network-operator.v4.7.0.clusterserviceversion.yaml
- sriov-network-operator-sriovnetwork.crd.yaml
- image-references 

Signed-off-by: Alexey Roytman <roytman@il.ibm.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,6 @@ IMAGE_BUILD_OPTS?=
 DOCKERFILE?=Dockerfile
 
 CRD_BASES=./config/crd/bases
-MANIFESTS_4.7=./manifests/4.7
 
 export APP_NAME=sriov-network-operator
 APP_REPO=github.com/openshift/$(APP_NAME)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -101,7 +104,12 @@ uninstall: manifests kustomize
 
 # Generate manifests e.g. CRD, RBAC etc.
 manifests: controller-gen
-	$(CONTROLLER_GEN) $(CRD_OPTIONS) webhook paths="./..." output:crd:artifacts:config=config/crd/bases
+	$(CONTROLLER_GEN) $(CRD_OPTIONS) webhook paths="./..." output:crd:artifacts:config=$(CRD_BASES)
```
zshi-redhat, line 107: Might be good to make ocp version as a var so that it can be used for future release updates.
There was an example in 4.6 release: https://github.com/openshift/sriov-network-operator/blob/release-4.6/Makefile#L111

zshi-redhat:
> Automatically copies generated `yaml` files from `./config/crd/bases` to `./manifests/4.7`
> 
> **Note:** the `./manifests/4.7` dirictory conatins 3 additional files, I've not found thier sources, so they probably should be continue to be manually updated.
> The files are:
> 
>     * sriov-network-operator.v4.7.0.clusterserviceversion.yaml

This file is updated manually.
> 
>     * sriov-network-operator-sriovnetwork.crd.yaml

Looks like the crd.yaml for sriovnetwork uses a different name format than other crd files under /manifests/4.7
I think we need to change `sriov-network-operator-sriovibnetworks_crd.yaml` under `/manifest` to `sriov-network-operator-sriovibnetwork.crd.yaml`
> 
>     * image-references

This file is updated manually, it won't change if no new image is added.
>

pliurh:
@roytman From operator-sdk 1.0.0, a new `bundle` target was introduced in the Makefile. which shall be used to generate the operator bundle (all the files we need in the manifests directory). I suggest we use that framework to generate the operator bundle.

Makefile
roytman: Done, thanks

roytman:
> Looks like the crd.yaml for sriovnetwork uses a different name format than other crd files under /manifests/4.7
> I think we need to change `sriov-network-operator-sriovibnetworks_crd.yaml` under `/manifest` to `sriov-network-operator-sriovibnetwork.crd.yaml`

Added the `sriov-network-operator-sriovnetwork.crd.yaml` file copy.

roytman:
> 
> 
> @roytman From operator-sdk 1.0.0, a new `bundle` target was introduced in the Makefile. which shall be used to generate the operator bundle (all the files we need in the manifests directory). I suggest we use that framework to generate the operator bundle.

Hi @pliurh, using the `generate bundle` is a more drastical change, in this PR I just wanted to reduce number of manual operations.

pliurh:
@roytman The header of the CRD file needs to be removed for the bundle files. PTAL https://github.com/openshift/sriov-network-operator/pull/398

roytman:
> 
> 
> @roytman The header of the CRD file needs to be removed for the bundle files. PTAL #398

Done.

PR #422, opened 2020-11-26
Bug 1901909: do not set the node selector in the dp / cni daemons if empty in config.

fedepaol:
The reconciliation loop of the policy sets the selector to worker/linux if the one in the config is empty. 
The config reconciliation, on the other hand, sets it to empty. When this happens, a new reconciliation is triggered and the pdos are recreated.

This causes a restart of the dp / cni pods every 5 minutes


fedepaol:
I think this is the cause for https://github.com/openshift/sriov-network-operator/issues/394

PR #423, opened 2020-11-27
Fix infiniband rendering

fedepaol:
The template now has a MetaPluginsConfigured plugin that is not set in case of infiniband. We set it to false as we don't expect metaplugins to be needed in case of ib.


zshi-redhat:
looks good to me

PR #424, opened 2020-11-27
Make the clean policies the last operation of clean.Clean.

fedepaol:
This allows clients to call sriov stable even in the single node scenario.
By having the policy cleanup the last operation of Clean, we ensure we are able to perform the other cleanings before the node is rebooted, avoiding unwanted errors.


SchSeba (approved):
/lgtm

PR #425, opened 2020-11-29
Fix logic in case SKIP_VAR_SET is set

oshoval:
If `SKIP_VAR_SET` is set, it should check the SRIOV_* env vars
because those are the ones that the user should override,
and not the DIGEST* ones, which are just intermediate values.


oshoval:
Thanks,
I believe the flakes are not due to this PR?

PR #426, opened 2020-11-29
Add support for link types selector

### LINKED ISSUE #427: Support link type selector for deploy device plugin
Currently, the operator doesn't support determining a specific link type NICs for the deployed device plugin e.g. determining a node to be only ethernet or InfiniBand NICs only.

Mmduh-483:
Currently, the operator has no way to determine a specific link type NICs for the deployed device plugin e.g. determining a node to be only ethernet or InfiniBand NICs only.

This patch adds support for "linkTypes" selector of SR-IOV network device plugin, which provides a more flexible selection of the NICs.

fixes #427


zshi-redhat:
@Mmduh-483 would it be sufficient to use rootDevices or pfNames to uniquely identify a node device already? Although it's not called out explicitly, I think we shall keep the user-facing API be simple as long as it can satisfy the need of selecting devices.

Mmduh-483:
@zshi-redhat In case the rootDevices or PFNames were used in a higher priority policy and link-type got changed, it the functionality, we need a more robust solution, WDYT?

Mmduh-483:
@pliurh @zshi-redhat can you please review the PR

PR #428, opened 2020-11-30
Only use env variable ENABLE_ADMISSION_CONTROLLER for initialization

pliurh:


controllers/sriovoperatorconfig_controller.go
```diff
@@ -95,16 +95,20 @@ func (r *SriovOperatorConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Result
 		return reconcile.Result{}, nil
 	}
 
-	if enableAdmissionController {
-		// Render and sync webhook objects
-		if err = r.syncWebhookObjs(defaultConfig); err != nil {
-			return reconcile.Result{}, err
-		}
+	// Render and sync webhook objects
+	if err = r.syncWebhookObjs(defaultConfig); err != nil {
```
zshi-redhat, line 99: Would this fail in non-openshift environment when calling `r.deleteWebhookObject`?

controllers/sriovoperatorconfig_controller.go
```diff
@@ -95,16 +95,20 @@ func (r *SriovOperatorConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Result
...
+		return reconcile.Result{}, err
+	}
 
-		// Render and sync CA configmap
+	if *defaultConfig.Spec.EnableInjector {
```
zshi-redhat, line 103: Add a note that in k8s environment, this should never set to true?

controllers/sriovoperatorconfig_controller.go
pliurh: No, it would not. The function only handles k8s objects. Those are common in openshift and vanilla k8s.

controllers/sriovoperatorconfig_controller.go
pliurh: In the switchdev PR, a CLUSTER_TYPE environment variable will be added to tell what platform the operator is running on. We can add that condition check after that PR is merged.

pliurh:
cc @Mmduh-483

controllers/sriovoperatorconfig_controller.go
zshi-redhat: We already using ENABLE_ADMISSION_CONTROLLER, why add CLUSTER_TYPE in the mix? plus in the future we may enable webhook for vanilla k8s.

PR #429, opened 2020-11-30
Fix tests of rendering net-att-def config

pliurh:


pliurh:
This shall be merged before https://github.com/openshift/release/pull/13934

Makefile
```diff
@@ -152,6 +152,9 @@ else
 KUSTOMIZE=$(shell which kustomize)
 endif
 
+skopeo:
```
fedepaol, line 155: What's the need for skopeo here?

pliurh: skopeo is required by hack/env.sh

zshi-redhat: Just in case, I think we don't assume local installation will work in a disconnected environment, right?

Issue #421, opened 2020-11-26
SriovNetworkNodeState status shows non-sriov interfaces


yoavzai:
It is understood from the documentation of SriovNetworkNodeState resource, that its status should show only PF and VF but in reality it shows not only sriov nics. I would suggest to note that in the documentation.

PR #430, opened 2020-11-30
Fix DiscoverSriovDevices to ignore non SRIOV PF interfaces

oshoval:
Current `DiscoverSriovDevices` logic doesn't ignore non SRIOV nics.

Fixes: https://github.com/openshift/sriov-network-operator/issues/421

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -66,6 +66,10 @@ func DiscoverSriovDevices() ([]sriovnetworkv1.InterfaceExt, error) {
 			continue
 		}
 
+		if !dputils.IsSriovPF(device.Address) {
+			continue
+		}
+
 		driver, err := dputils.GetDriverName(device.Address)
 		if err != nil {
 			glog.Warningf("DiscoverSriovDevices(): unable to parse device driver for device %+v %q", device, err)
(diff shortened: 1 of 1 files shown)


pkg/utils/utils.go
```diff
@@ -66,6 +66,10 @@ func DiscoverSriovDevices() ([]sriovnetworkv1.InterfaceExt, error) {
 			continue
 		}
 
+		if !dputils.IsSriovPF(device.Address) {
```
zshi-redhat, line 69: This may not be accurate for mellanox NIC, which requires configuring firmware to enable sriov in mellanox_plugin.

@adrianchiris is there a reliable way to check whether a device is capable of SR-IOV when ENABLE_SRIOV is false in firmware?

adrianchiris: In general, all Nvidia NIC hardware is SR-IOV capable.
However if its disabled in Firmware via non-volatile configurations  (mstconfig) it is not advertised to the kernel
(requires SRIOV_EN=true and NUM_OF_VFS>0 in firmware non-volatile configurations)

no reliable way to check except to assume its supported if pci vendor ID is `15b3`

when either: `SRIOV_EN=false` or `NUM_OF_VFS=0` is set. `sriov_numvfs`, `sriov_totalvfs` files will not exist

adrianchiris: currently this patch will not discover any Nvidia NIC PFs if SRIOV is not configured

oshoval: Ok thanks, so closing the PR as was suggested by @zshi-redhat

oshoval: do we want to check the vendor ID as you suggested?
and then i will reopen the PR and update it.

adrianchiris: regardless of Nvidia NICs, 
I think It will not work in a virtualized environment where VFs are passthrough to the VM.

zshi-redhat: VF discovery in VM is handled separately: https://github.com/openshift/sriov-network-operator/blob/master/pkg/utils/utils_virtual.go#L15

zshi-redhat: If we were to add vendor ID check, the code might belong to vendor specific plugin. For example, each plugin provides a function that will be called by `DiscoverSriovDevices` to determine whether the device is sriov capable. For mellanox plugin, it can just return true meaning every mellanox NIC supports SRIOV. For other plugin, they can check the device ID and return accordingly.

oshoval: Thank you both, so as we spoke i am closing this PR,
so the generalized solution will be implemented.

PR #431, opened 2020-11-30
Bug 1900736: Use node specific data for device plugin config

zshi-redhat:


markmc:
(patch manager) noting that openshift/sriov-network-device-plugin#35 is also required. Fixes a serious issue encountered by a real-world user. Clean backport from master.

PR #434, opened 2020-12-02
Fix metaplugins name to metaPlugins to meet naming convention

Mmduh-483:


zshi-redhat:
Shall we change the API definition from `metaplugins` to `metaPlugins`? which will align with naming conventions of other fields.

Mmduh-483:
> Shall we change the API definition from `metaplugins` to `metaPlugins`? which will align with naming conventions of other fields.

@zshi-redhat fixed

pliurh:
> > Shall we change the API definition from `metaplugins` to `metaPlugins`? which will align with naming conventions of other fields.
> 
> @zshi-redhat fixed

CRD files in manifests/4.7 shall be updated as well.

Mmduh-483:
> CRD files in manifests/4.7 shall be updated as well.

@zshi-redhat it is already `metaPlugins` 
https://github.com/openshift/sriov-network-operator/blob/master/manifests/4.7/sriov-network-operator-sriovnetwork.crd.yaml#L61

PR #443, opened 2020-12-14
Fixed findUnusedSriovDevices function to Skip tests

kononovn:
Fixed findUnusedSriovDevices function to Skip tests instead raise Error.
Allow to use DOWN interfaces for tests.

### DIFF
--- a/test/conformance/tests/sriov_operator.go
+++ b/test/conformance/tests/sriov_operator.go
@@ -1082,7 +1082,9 @@ var _ = Describe("[sriov] operator", func() {
 					sriovDeviceList, err := sriovInfos.FindSriovDevices(testNode)
 					Expect(err).ToNot(HaveOccurred())
 					unusedSriovDevices, err := findUnusedSriovDevices(testNode, sriovDeviceList)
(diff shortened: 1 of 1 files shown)


SchSeba:
/hold

@fedepaol can you have a quick look?

test/conformance/tests/sriov_operator.go
```diff
@@ -1604,17 +1604,14 @@ func findUnusedSriovDevices(testNode string, sriovDevices []*sriovv1.InterfaceEx
...
-		}
 		if strings.Contains(stdout, "master ovs-system") {
 			continue // The interface is not active
 		}
 
 		filteredDevices = append(filteredDevices, device)
 	}
 	if len(filteredDevices) == 0 {
-		return nil, fmt.Errorf("Unused sriov devices not found")
+		Skip("Unused sriov devices not found")
```
fedepaol, line 1614: Can you handle the error and skip from outside?
We don't know how this is gonna be used, and I'd prefer to keep it independent from the test behaviour.

SchSeba (approved):
/lgtm
/approve

PR #445, opened 2020-12-14
Support SKIP_VAR_SET in Makefile

oshoval:
In case a non empty `SKIP_VAR_SET` is exported,
Makefile would respect it and not install `skopeo`
in case `skopeo` doesn't exists.

This will allow to use the sriov-operator with
non root users.

See please https://github.com/openshift/sriov-network-operator/blob/52480752f6582cf2829845f291a147d75c23756e/hack/env.sh#L1
for more details about `SKIP_VAR_SET` and its purpose to allow
running without `skopeo` installed, when env vars are overridden by the user.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -170,7 +170,7 @@ KUSTOMIZE=$(shell which kustomize)
 endif
 
 skopeo:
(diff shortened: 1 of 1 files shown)


oshoval:
If needed will move to https://github.com/k8snetworkplumbingwg/sriov-network-operator

oshoval:
Opened same on upstream, if needed will open here once its merged on upstream.

PR #447, opened 2020-12-14
Bug 1898594: Fix the sriov daemon selector assignement

fedepaol:

This is a manual cherry pick of https://github.com/openshift/sriov-network-operator/pull/413


fedepaol:
@zshi-redhat @pliurh PTAL

PR #475, opened 2021-02-22
Bug 1879379: Change must-gather/Dockerfile.rhel7 to use must-gather image

oribon:
This should fix the problem of oc adm must-gather not finding oc binary


zshi-redhat:
We probably don't need this PR for upstream repo for now, LGTM.

/lgtm

PR #478, opened 2021-02-23
Increase tests timeout for single node test environments

mmirecki:
Single node environments can take much longer to stabilize than environment with multiple nodes. This can cause existing tests to fail due to the timeouts not being long enough. This PR increases the timeouts if a single node environment is detected.


SchSeba (approved):
/lgtm
/approve

PR #479, opened 2021-02-25
[release-4.7] Change must-gather/Dockerfile.rhel7 to use must-gather image

oribon:
This should fix the problem of oc adm must-gather not finding oc binary


pliurh:
LGMT
/cc @zshi-redhat

PR #490, opened 2021-04-08
Add management workload annotations

lack:
In support of the workload partitioning feature
(https://github.com/openshift/enhancements/pull/703), we need to add
annotations to all management pods and namespaces so they can be
properly identified and assigned to segregated management cores on
clusters configured to do so.

### DIFF
--- a/bindata/manifests/daemon/daemonset.yaml
+++ b/bindata/manifests/daemon/daemonset.yaml
@@ -14,6 +14,8 @@ spec:
     type: RollingUpdate
   template:
     metadata:
+      annotations:
+        workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
       labels:
         app: sriov-network-config-daemon
         component: network
--- a/bindata/manifests/operator-webhook/server.yaml
+++ b/bindata/manifests/operator-webhook/server.yaml
@@ -18,6 +18,8 @@ spec:
     type: RollingUpdate
   template:
     metadata:
+      annotations:
+        workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
       labels:
         app: operator-webhook
     spec:
--- a/bindata/manifests/plugins/sriov-cni.yaml
+++ b/bindata/manifests/plugins/sriov-cni.yaml
@@ -16,6 +16,8 @@ spec:
     type: RollingUpdate
   template:
     metadata:
+      annotations:
+        workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
       labels:
         app: sriov-cni
         component: network
--- a/bindata/manifests/plugins/sriov-device-plugin.yaml
+++ b/bindata/manifests/plugins/sriov-device-plugin.yaml
@@ -16,6 +16,8 @@ spec:
     type: RollingUpdate
   template:
     metadata:
+      annotations:
+        workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
       labels:
         app: sriov-device-plugin
         component: network
--- a/bindata/manifests/webhook/server.yaml
+++ b/bindata/manifests/webhook/server.yaml
@@ -18,6 +18,8 @@ spec:
     type: RollingUpdate
   template:
     metadata:
+      annotations:
+        workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
       labels:
         app: network-resources-injector
         component: network
--- a/manifests/4.8/sriov-network-operator.v4.8.0.clusterserviceversion.yaml
+++ b/manifests/4.8/sriov-network-operator.v4.8.0.clusterserviceversion.yaml
@@ -297,6 +297,8 @@ spec:
               name: sriov-network-operator
           template:
             metadata:
+              annotations:
+                workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
               labels:
                 name: sriov-network-operator
             spec:


zshi-redhat:
@lack thanks for sending the PR!

Several questions and comments:
1) Do we need this PR for 4.8 FF?
2) There is an [upstream project](https://github.com/k8snetworkplumbingwg/sriov-network-operator) for sriov-network-operator where you might need to send the PR to, then we will backport the change.
3) CLUSTER_TYPE` env var is used in upstream project to differentiate an OpenShift and vanilla K8s cluster, you would need to add a conditional check for the CLUSTER_TYPE in the bindata manifests and other places where appropriate.

Let us know if you need any help.

lack:
> 1. Do we need this PR for 4.8 FF?

Yes!  We know it's tight, and may be applying for exceptions due to its high-priority status.

> 2. There is an [upstream project](https://github.com/k8snetworkplumbingwg/sriov-network-operator) for sriov-network-operator where you might need to send the PR to, then we will backport the change.

Good to know, I'll check it out.

> 3. CLUSTER_TYPE` env var is used in upstream project to differentiate an OpenShift and vanilla K8s cluster, you would need to add a conditional check for the CLUSTER_TYPE in the bindata manifests and other places where appropriate.

It's also harmless for these annotations to be set on non-OpenShift clusters; they'll just be ignored.

I'll take a look at getting this in upstream and let you know if I have questions!  Thanks @zshi-redhat !

lack:
@zshi-redhat : So what would you prefer for the upstream change?  Just adding these annotations indescriminately?   I didn't see any obvious point where they could be made optional based on the install type, and there's no harm in having these annotations on non-OpenShift clusters.

zshi-redhat:
> @zshi-redhat : So what would you prefer for the upstream change? Just adding these annotations indescriminately? I didn't see any obvious point where they could be made optional based on the install type, and there's no harm in having these annotations on non-OpenShift clusters.

@lack 
I was thinking the change would be something like this:
Add the CLUSTER_TYPE env var in Operator [manifests](https://github.com/openshift/sriov-network-operator/blob/master/manifests/4.8/sriov-network-operator.v4.8.0.clusterserviceversion.yaml#L316)
```
                  env:
                    - name: CLUSTER_TYPE
                      value: "openshift"
```

Then, in Operator controller [code](https://github.com/openshift/sriov-network-operator/blob/9ed7e5c8bfdfa3fd0b8ecce145aa44a079b19ac4/controllers/sriovnetworknodepolicy_controller.go#L320), use the env var to render operator managed manifests:
```
data.Data["ClusterType"] = os.Getenv("CLUSTER_TYPE")
```

Then, in the bindata, e.g device-plugin [daemon](https://github.com/openshift/sriov-network-operator/blob/9ed7e5c8bfdfa3fd0b8ecce145aa44a079b19ac4/bindata/manifests/plugins/sriov-device-plugin.yaml). 

```
If  {{.ClusterType}} == openshift
    workload management annotation
{{end}}

```
Thinking this again, we perhaps don't want to make all these changes just for adding the annotations (in upstream for openshift).
After talking with Peng, we agreed to having this PR downstream only.

Sorry for the misleading comments!

Let's review this PR and get it merged.

config/default/manager_auth_proxy_patch.yaml
```diff
@@ -7,6 +7,9 @@ metadata:
   namespace: system
 spec:
   template:
+    metadata:
```
zshi-redhat, line 10: I don't think we use this config for generating openshift 4.8 manifest.  /cc @pliurh 
The change is not needed.

config/default/manager_webhook_patch.yaml
```diff
@@ -5,6 +5,9 @@ metadata:
   namespace: system
 spec:
   template:
+    metadata:
```
zshi-redhat, line 8: ditto

config/manager/manager.yaml
```diff
@@ -1,6 +1,8 @@
 apiVersion: v1
 kind: Namespace
 metadata:
+  annotations:
```
zshi-redhat, line 4: ditto

deploy/namespace.yaml
```diff
@@ -2,5 +2,7 @@ apiVersion: v1
 kind: Namespace
 metadata:
   name: $NAMESPACE
+  annotations:
```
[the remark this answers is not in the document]
pliurh: Also, why would we annotate the namespace? We asked users to create the namespace manually. Do we need to change the document to add the annotation?

config/default/manager_auth_proxy_patch.yaml
pliurh: Correct.

deploy/namespace.yaml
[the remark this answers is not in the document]
lack: You're right!

The namespace does need this annotation, but for an operator like this, that's up to the administrator.

For our use of this operator we have some additional scripting that creates the namespace and will add the annotation there.  For general use, we will have documentation about this.

zshi-redhat:
Current changes look good to me.
We would need a linked BZ to merge. @lack would you mind creating one?

deploy/namespace.yaml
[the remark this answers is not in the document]
zshi-redhat: Besides adding in the official doc, maybe also change the creating of namespace from cmd to yaml (like how it is done in ptp-operator: https://github.com/openshift/ptp-operator/pull/106/files#diff-6317830c9ed3d65ddc9b69b7533adc00c92892735761228723d1a9f9fc3d1a15R110)

lack:
> Current changes look good to me.
> We would need a linked BZ to merge. @lack would you mind creating one?

That's in the works now (for ~30 other PRs too of the same flavor).  I'll update with the actual issue as soon as it's available!

lack:
@zshi-redhat : bugzilla approval achieved!  Ready for final approval and merge.

zshi-redhat:
@lack would you please link the BZ in this PR?

/lgtm

lack:
@zshi-redhat : Certainly!  https://bugzilla.redhat.com/show_bug.cgi?id=1949598

PR #499, opened 2021-05-08
Bug 1955874: Sync upstream: 2021-05-08

zshi-redhat:


zshi-redhat:
note:  link the bug that will be created from https://bugzilla.redhat.com/show_bug.cgi?id=1870547#c9.

PR #500, opened 2021-05-08
Bug 1958467: Fix index out of range panic when ifaceName is empty

zshi-redhat:
Signed-off-by: Zenghui Shi <zshi@redhat.com>


mrunalp:
The tests need to be fixed here. Removing cherry-pick-approved.

PR #504, opened 2021-05-13
[release-4.7] Bug 1960103: Pause MCP before draining/rebooting node

pliurh:
The config daemon instance on each node need to compete for the 'Draining' lock before
it can draining/rebooting a node.

After a config daemon instance gets the lock, it checks its MachineConfigPool conditions.
If the MCP is in ready state, it will pause the MCP then process, otherwise, it wait for
the MCP getting ready. The MCP will be resumed after the config daemon and release the
lock after the config dameon finishes its work.


zshi-redhat:
@pliurh is this a clean merge?

pliurh:
> @pliurh is this a clean merge?

No, there were some conflicts during cherry-pick. I resolved them manual

PR #513, opened 2021-06-02
Added reviewer nicklesimba to OWNERS

nicklesimba:
As titled


pliurh:
@nicklesimba It looks like you need to become a member of openshift org first.

PR #516, opened 2021-06-16
Exclude non yamls from bundleci.Dockerfile

oribon:
ci-operator does not support ImageStream resources


fedepaol:
/lgtm

@zshi-redhat PTAL

PR #532, opened 2021-06-30
Add infrastructure feature annotation in operator CSV

pliurh:


pliurh:
merged this fix with #543

zshi-redhat:
> merged this fix with #543

How does #543 cover this PR?

PR #533, opened 2021-06-30
Fix daemon pod log

sabinaaledort:
Some spaces were missing and the namespace should be before the pod name.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -167,8 +167,7 @@ func New(
 				if usingEviction {
 					verbStr = "Evicted"
 				}
-				glog.Info(fmt.Sprintf("%s pod from Node", verbStr),
-					"pod", fmt.Sprintf("%s/%s", pod.Name, pod.Namespace))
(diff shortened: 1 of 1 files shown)


pkg/daemon/daemon.go
```diff
@@ -168,7 +168,7 @@ func New(
 					verbStr = "Evicted"
 				}
 				glog.Info(fmt.Sprintf("%s pod from Node", verbStr),
-					"pod", fmt.Sprintf("%s/%s", pod.Name, pod.Namespace))
+					" pod ", fmt.Sprintf("%s/%s", pod.Namespace, pod.Name))
```
SchSeba, line 171: maybe even remove the pod here?

also mabe have one fmt.Sprintf

SchSeba:
@sabinaaledort please open this PR against the https://github.com/k8snetworkplumbingwg/sriov-network-operator

sabinaaledort:
This change was done in the u/s repo and will be copied in the next sync from upstream.
https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/154

PR #536, opened 2021-07-01
Bug 1978435: Add labels to indicate availability of this operator for ppc64le

manojnkumar:
This was incorrectly being set in the spec.labels.


yselkowitz:
This is going to have to be manually back-ports
/cherry-pick cancel

PR #537, opened 2021-07-01
Bug 1978437: Add labels to indicate availability of this operator for ppc64le

manojnkumar:
This was incorrectly set in spec.labels


manojnkumar:
Aborting.

PR #542, opened 2021-07-14
Sync upstream: 2021-7-14

pliurh:


zshi-redhat:
@pliurh we add the bundle folder for openshift only, right?
I didn't find the same upstream. do you see us updating upstream repo as well?

pliurh:
Yes, this is only for downstream. In the operator-sdn upgrade PR, I removed the `bundle` directory from the repo. As we don't use the operator-sdk to generate the operator bundle. However, in the PR, I found our CI bundle pipeline requires the existence of `bundle\metadate` directory. So I added that fold back. 

BTW, I'm working on another PR that allows us to use the operator-sdk to generate the bundle.

PR #547, opened 2021-07-29
Added Intel device 1572

evgenLevin:


SchSeba:
Hi @zshi-redhat I open this one instead

https://github.com/openshift/sriov-network-operator/pull/548

the question is how we can maybe skip the BZ validation as this only effect the tests

zshi-redhat:
> Hi @zshi-redhat I open this one instead
> 
> #548
> 
> the question is how we can maybe skip the BZ validation as this only effect the tests

I assume you can create a bz for the backport or maybe override the bugzilla tag in the PR (I didn't try the latter).

SchSeba:
@zshi-redhat I don't have permission to manually add the BZ labels do you have?

The problem with the BZ is we will need them all the way from 4.9 to 4.7 and PRs that are already merge a long time ago

zshi-redhat:
> @zshi-redhat I don't have permission to manually add the BZ labels do you have?
> 
> The problem with the BZ is we will need them all the way from 4.9 to 4.7 and PRs that are already merge a long time ago

I usually create the bug all the way for master to released branch and set the bug to Verified state for those already merged/verified.  I can overwrite the bugzilla/invalid-bug, but we need z-stream manager to approve the PR..

PR #550, opened 2021-08-04
Bug 1989456: Remove ServiceAccount sriov-network-operator from operator bundle

pliurh:


bundle/manifests/sriov-network-operator_v1_serviceaccount.yaml
```diff
@@ -1,5 +0,0 @@
-apiVersion: v1
-kind: ServiceAccount
-metadata:
-  creationTimestamp: null
-  name: sriov-network-operator
```
timflannagan, line 5: :+1:

config/rbac/service_account.yaml
```diff
@@ -1,4 +0,0 @@
-apiVersion: v1
-kind: ServiceAccount
-metadata:
-  name: sriov-network-config-daemon
```
timflannagan, line 4: I have zero context on your bundle contents, but had a quick question - This ServiceAccount manifest metadata.Name wouldn't match the `serviceAccountName` specified in the CSV manifest, so deleting this may not be needed. It's possible I'm missing something and this was renamed to `sriov-network-config-daemon_service_account.yaml` like I'm seeing in the kustomize.yaml changes but I didn't see that change in this PR.

pliurh: I found we have `sriov-network-config-daemon_service_account.yaml` with identical content. So I remove this file and modify the kustomization file.  This is just a cleanup.

timflannagan: Thanks for confirming that - this lgtm.

PR #588, opened 2021-11-25
Add Nvidia device ID 15b3/101f/101e

zshi-redhat:
Signed-off-by: Zenghui Shi <zshi@redhat.com>


SchSeba:
should we also open a PR to change it here?

https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/deploy/configmap.yaml#L19

zshi-redhat:
> should we also open a PR to change it here?
> 
> https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/deploy/configmap.yaml#L19

We take different approach tracking supported NICs downstream.

PR #604, opened 2021-12-16
Bug 2028531: [cherry-pick from upstream] validate: allow netFilter to be set alone in NicSelector

EmilienM:
In staticValidateSriovNetworkNodePolicy, we now allow a NicSelector to
only have NetFilter defined.

This will be useful for the OpenStack platform, where we can request a
Network Node Policy only by its Network ID.

### DIFF
--- a/pkg/webhook/validate.go
+++ b/pkg/webhook/validate.go
@@ -3,14 +3,15 @@ package webhook
 import (
 	"context"
 	"fmt"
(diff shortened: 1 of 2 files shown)


pliurh:
Close this PR as we normally do patch sync for master.
https://github.com/openshift/sriov-network-operator/pull/605

PR #606, opened 2021-12-21
BUG 2013199: Update operator bundle

pliurh:


manifests/4.10/supported-nic-ids_v1_configmap.yaml
```diff
@@ -22,8 +22,6 @@ data:
   Nvidia_mlx5_ConnectX-5_Ex: 15b3 1019 101a
   Nvidia_mlx5_ConnectX-6: 15b3 101b 101c
   Nvidia_mlx5_ConnectX-6_Dx: 15b3 101d 101e
-  Nvidia_mlx5_ConnectX-6_Lx: 15b3 101f 101e
```
zshi-redhat, line 25: Why removing these two?

pliurh: It was synced from upstream. I think we shall keep it unchanged.

pliurh: fixed

zshi-redhat:
we would also need this change in file `deployment/sriov-network-operator/crds/sriovnetwork.openshift.io_sriovnetworknodestates.yaml`?

pliurh:
> we would also need this change in file `deployment/sriov-network-operator/crds/sriovnetwork.openshift.io_sriovnetworknodestates.yaml`?

Yes, but not in this PR. That file is only used by the upstream installation.

zshi-redhat:
> > we would also need this change in file `deployment/sriov-network-operator/crds/sriovnetwork.openshift.io_sriovnetworknodestates.yaml`?
> 
> Yes, but not in this PR. That file is only used by the upstream installation.

Do we already have an upstream PR for this change?

PR #613, opened 2022-01-29
Sync deploy/configmap.yaml with 4.10 supported nics

EmilienM:
This will help to use this configmap with support nics in our CI.

### DIFF
--- a/deploy/configmap.yaml
+++ b/deploy/configmap.yaml
@@ -1,22 +1,32 @@
 apiVersion: v1
+data:
+  Broadcom_bnxt_en_BCM57414_NetXtreme-E: 14e4 16d7 16c1
+  Broadcom_bnxt_en_BCM57508_NetXtreme-E: 14e4 1750 1806
+  Intel_i40e_X710_10G: 8086 1572 154c
+  Intel_i40e_XL710_40G: 8086 1583 154c
+  Intel_i40e_XXV710: 8086 158a 154c
+  Intel_i40e_XXV710_25G: 8086 158b 154c
+  Intel_i40e_XXV710_N3000: 8086 0d58 154c
+  Intel_ice_E810-CQDA2_2CQDA2: 8086 1592 1889
+  Intel_ice_E810-XXVDA2: 8086 159b 1889
+  Intel_ice_E810-XXVDA4: 8086 1593 1889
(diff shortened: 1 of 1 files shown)


SchSeba:
Hi @EmilienM adding these nics into the list will mean we fully support these nics and I don't think that is the case or what we want to achieve here.

will it be possible as part of your CI system to just add these nics into the configmap after you deploy the operator I think that should be possible.

EmilienM:
> Hi @EmilienM adding these nics into the list will mean we fully support these nics and I don't think that is the case or what we want to achieve here.

Well, the additions in this patch was just a copy paste from `manifests/4.10/supported-nic-ids_v1_configmap.yaml` which _is_ supported. So I don't understand what the issue is. These NICs are already supported, isn't?

> will it be possible as part of your CI system to just add these nics into the configmap after you deploy the operator I think that should be possible.

Yes I can hack but this wasn't my goal.

SchSeba:
Hi @EmilienM I was not aware this was already in the configmap sorry about that :)

can you please open this PR against the u/s repo?
The bundle folder doesn't exist there but the deploy is.

EmilienM:
> Hi @EmilienM I was not aware this was already in the configmap sorry about that :)
> 
> can you please open this PR against the u/s repo? The bundle folder doesn't exist there but the deploy is.
> 
> https://github.com/k8snetworkplumbingwg/sriov-network-operator

upstream doesn't have `manifests/4.10/supported-nic-ids_v1_configmap.yaml` file, which is why I didn't push it upstream.
Do you want me to sync `manifests/4.10/supported-nic-ids_v1_configmap.yaml` content from downstream to the upstream configmap used in `deploy/` directory?

EmilienM:
any update on this one?

EmilienM:
@SchSeba would you please looking at it again please? Thanks

pliurh:
We have different support NIC list at u/s and d/s. The file in `manifests/4.10/supported-nic-ids_v1_configmap.yaml` will be deployed as part of the operator bundle at d/s.  The file `deploy/configmap.yaml` is only used for local development, so I think it is ok to modify it at both u/s and d/s.

EmilienM:
this PR hasn't got traction, I'll abandon it.

PR #618, opened 2022-02-02
Bug 2049774: Fix webhook for virtual workers

EmilienM:
This is a manual backport (was in cherry-pick conflict otherwise).

This commit fix the validation webhook to support virtual workers using
PCI passthrough


EmilienM:
@SchSeba FYI

PR #637, opened 2022-03-02
bug 2059292: Fix error when deploying sriov network operator

bn222:
https://bugzilla.redhat.com/show_bug.cgi?id=2059292

### DIFF
--- a/bundle.Dockerfile
+++ b/bundle.Dockerfile
@@ -6,7 +6,7 @@ LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
 LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
 LABEL operators.operatorframework.io.bundle.package.v1=sriov-network-operator
 LABEL operators.operatorframework.io.bundle.channels.v1=alpha
-LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.9.0+git
+LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.16.0+git
 LABEL operators.operatorframework.io.metrics.mediatype.v1=metrics+v1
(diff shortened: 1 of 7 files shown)


zshi-redhat:
@bn222 We would need a 4.10 and master BZs (if not already exist) in verified state in order for this PR to be labeled with bugzilla/valid-bug.

manifests/4.9/supported-nic-ids_v1_configmap.yaml
```diff
@@ -22,7 +22,6 @@ data:
   Nvidia_mlx5_ConnectX-5_Ex: 15b3 1019 101a
   Nvidia_mlx5_ConnectX-6: 15b3 101b 101c
   Nvidia_mlx5_ConnectX-6_Dx: 15b3 101d 101e
-  Nvidia_mlx5_ConnectX-6_Lx: 15b3 101f 101e
```
SchSeba, line 25: is there a reason why we remove this device?

SchSeba:
@bn222 is the change exists in 4.10 I was able to find it.

we are also deploying the sriov operator as part of the CI and didn't encounter this issue.

manifests/4.9/supported-nic-ids_v1_configmap.yaml
bn222: Same as here: https://github.com/openshift/sriov-network-operator/pull/639#discussion_r833046689

bn222:
> @bn222 is the change exists in 4.10 I was able to find it.
> 
> we are also deploying the sriov operator as part of the CI and didn't encounter this issue.

That's a bit unexpected. On a clean 4.9 env using dev-scripts, I can not deploy the sriov network operator. Any ideas if the CI is not caching anything from previous runs?

bn222:
Related bug has been closed with WORKSFORME

PR #639, opened 2022-03-08
Bug 2060322: Add RBAC for 'infrastructures' to operator bundle  @pliurh @bn222

bn222:


bn222:
@zshi-redhat

manifests/4.10/supported-nic-ids_v1_configmap.yaml
```diff
@@ -22,8 +22,6 @@ data:
   Nvidia_mlx5_ConnectX-5_Ex: 15b3 1019 101a
   Nvidia_mlx5_ConnectX-6: 15b3 101b 101c
   Nvidia_mlx5_ConnectX-6_Dx: 15b3 101d 101e
-  Nvidia_mlx5_ConnectX-6_Lx: 15b3 101f 101e
```
pliurh, line 25: do we want to remove these NICs from the supported list?

bn222: hm thanks for catching this. That was not the intention.

bn222: This happens when I run `make bundle`.

The device `Nvidia_mlx5_ConnectX-6_Lx` is missing from `config/manifests/bases/sriov-network-operator_configmap.yaml`.

It is not clear to me if it should be added to that yaml file, or removed by `make bundle`.

pliurh: We maintain different support nic list between u/s and d/s. So we shall update `make bundle` to ignore this file. For this PR, I think you can just remove the change to this file.

fabrizio8 (approved):
LGTM

PR #648, opened 2022-04-05
[release-4.10] Bug 2072038: backport leader election for sno to 4.10

eggfoobar:


SchSeba:
Hi @eggfoobar the PR is failing because the RBAC is not part of the CSV. 
you need to configure the RBAC in config/rbac folder and then generate the bundle that should add them to the CSV file

SchSeba:
@zshi-redhat @pliurh can you please add the label I don't have permission to do it

eggfoobar:
@SchSeba is there anything else missing for this PR to go through?

PR #651, opened 2022-04-06
Bug 2072498: fix make bundle

bn222:
Don't overwrite the supported NICs downstream. We maintain a different
list of NICs than upstream does.


bn222:
@SchSeba

PR #659, opened 2022-05-06
Update Makefile for go 1.18 builds

wizhaoredhat:
In go 1.18, the "go get" command no longer installs the module. This
effectively makes every "go get" command having the "-d" option turned
on implicitly. The solution is to now explicitly add "-d" to "go get"
for backwards compatibility with go versions <1.18. Then the module
is installed by the "go install" following step.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -129,12 +129,16 @@ generate: controller-gen
 	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
 
 CONTROLLER_GEN = $(shell pwd)/bin/controller-gen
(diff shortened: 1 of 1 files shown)


pliurh:
Normally, we follow operator-sdk's [guide](https://sdk.operatorframework.io/docs/upgrading-sdk-version/version-upgrade-guide/) to update the Makefile. Currently the operator-sdk is still with go 1.17. So I think we may want to wait until https://github.com/operator-framework/operator-sdk/issues/5481 closes.

BTW, the we haven't upgrade the operator-sdk since version 1.10, I suggest we follow [guide](https://sdk.operatorframework.io/docs/upgrading-sdk-version/version-upgrade-guide/) to upgrade the operator-sdk to the latest for this project first.

PR #671, opened 2022-06-23
Generates random MAC address when effective virtual function MAC addr…

vrindle:
…ess is zero.

This is to set VF effective MAC addresses when the VF effective MAC address is all zero for some vendor NICs.


SchSeba:
The PR on u/s was reverted I will like to not have d/s only changes..

SchSeba:
closing this as we have a new PR
https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/328

/close

PR #683, opened 2022-08-10
ensure network daemon checks if machineConfigPool resource registered before attempting to pause the MachineConfigPool

relyt0925:
In hypershift: there are no machineConfigPool resources registered and therefore no need to pause the MachineConfigPool resource when sriov-network-operator is orchestrating sr-iov configuration. This adds a check to see if that api resource is registered with the cluster before attempting to pause the machineConfigPool resource.

This would prevent the need from having to use CLUSTER_TYPE: "kubernetes" in hypershift clusters in order to get sriov-network-operator to work in those environments.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -785,10 +785,33 @@ func (dn *Daemon) getDrainLock(ctx context.Context, done chan bool) {
 	})
 }
 
+// isMachineConfigPoolResourceRegistered checks to see if the MachineConfigPool API Resource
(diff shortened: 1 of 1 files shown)


SchSeba:
Hi @relyt0925 please open the PR on the u/s repo https://github.com/k8snetworkplumbingwg/sriov-network-operator

Thanks!
/close

relyt0925:
Thank you for the guidance: PR is now here:
https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/350

PR #685, opened 2022-08-16
Updated golang.org/x/text to v0.3.7 Fixes Bug 2105595

vrindle:
Fixes this bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=2105595


wizhaoredhat:
FYI: Upstream has this commit that bumps golang.org/x/text to v0.3.7. https://github.com/openshift/sriov-network-operator/commit/38fb014397da62886de4ad2a07daaf6b288fa4ce#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6
In your case, I think just changing the golang.org/x/text should be fine as just a downstream commit on a release branch. Although I would be interested what the rules for this is.

vrindle:
> we already have that here also https://github.com/openshift/sriov-network-operator/blob/master/go.mod#L137
> 
> Am I missing something?

We realize the fix is in master, but it's not in 4.11. https://bugzilla.redhat.com/show_bug.cgi?id=2105596#c4

SchSeba:
Hi @vrindle don't worry I am going to do a backport of the 1.18 bump to 4.11 as OCP 4.11 also uses golang 1.18 that should fix the issue

SchSeba:
it will get fixed here https://github.com/openshift/sriov-network-operator/pull/688

SchSeba:
/close

This was fixed by the backport

PR #698, opened 2022-09-13
OCPBUGS-1235: Add missing Red Hat Virtio network device into supported nics

EmilienM:
We had this NIC supported in previous releases, and it was missed in 4.12 probably due to a bad copy/paste.
We need this NIC to be supported, for the virtualized environments (e.g. OpenStack).


SchSeba (approved):
/lgtm
/approve

PR #701, opened 2022-09-16
OCPVE-174: Support graceful node shutdown

pacevedom:
- Remove MachineConfig permissions
- Remove MachineConfigPool operations
- Add signal handler for SIGTERM
- Add an updating flag to signal when the plugins need to drain
- Add a terminationGracePeriodSeconds to config daemon

Open points:
- Prestop hooks. These get executed before SIGTERM is sent, what if we are in the middle of a plugin apply?
- Need to determine terminationGracePeriodSeconds by taking the longest plugin apply.
- Need help to test it.

/cc @SchSeba


pacevedom:
@SchSeba can you please have a look to see if everything falls into place, or did I miss something?
I am missing vendoring now, but I am expecting changes to the PR so I didnt do it yet.
/assign @SchSeba

pacevedom:
Opening in upstream.

PR #702, opened 2022-09-21
NHE-19: Support SR-IOV operator on HyperShift hosted cluster

bn222:
Allow to create sriov pods on workers and don't call into MCO if running on hypershift


bn222:
@zshi-redhat

SchSeba:
Hi @bn222 @zshi-redhat I may miss something but I am not sure this will work on hypershift as the operator still have the master selector (https://github.com/openshift/sriov-network-operator/blob/master/manifests/stable/sriov-network-operator.clusterserviceversion.yaml#L369)

deploy/operator.yaml
```diff
@@ -79,6 +79,10 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.name
+            - name: NODE_NAME
```
SchSeba, line 82: we need to sync this to the bundle and the manifests folder to have this part of OLM deployment

zshi-redhat:
> Hi @bn222 @zshi-redhat I may miss something but I am not sure this will work on hypershift as the operator still have the master selector (https://github.com/openshift/sriov-network-operator/blob/master/manifests/stable/sriov-network-operator.clusterserviceversion.yaml#L369)

@SchSeba You're right, we need to update downstream-only manifests under `manifests/stable` which is not part of the upstream hypershift change. @bn222 could you update?

bn222:
@zshi-redhat There was an offline discussion between me / @SchSeba  and some OLM folks. There should be a way to override this as you deploy using the operator hub. It's on my list to find the exact commands/steps (to be documented). Once we have more clarity on those steps, the comment from @SchSeba will be resolved.

bn222:
```
apiVersion: operators.coreos.com/v1alpha1
kind: Subscription
metadata:
  name: sriov-network-operator-subscription
  namespace: openshift-sriov-network-operator
spec:
  channel: "4.11"
  name: sriov-network-operator
  source: redhat-operators
  sourceNamespace: openshift-marketplace
  config:
    nodeSelector:
      node-role.kubernetes.io/worker: ""
```

PR #714, opened 2022-10-28
End2End - Pod Security Admission fixes

cgoncalves:
Sync of https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/368

These changes aim to solve [Pod Security Admission](https://kubernetes.io/docs/concepts/security/pod-security-admission/) problem when running tests against k8s v1.25+ , as privileged pods are often needed during tests.

### DIFF
--- a/test/util/clean/clean.go
+++ b/test/util/clean/clean.go
@@ -26,9 +26,7 @@ func All() error {
 			return fmt.Errorf("failed to restore node drain state %v", err)
 		}
 	}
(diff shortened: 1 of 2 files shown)


SchSeba:
Hi @cgoncalves thanks for the PR but for us to not lose track of changes will you be able to propose a full sync between the repos?

like this https://github.com/openshift/sriov-network-operator/pull/709

cgoncalves:
Sure. Here: https://github.com/openshift/sriov-network-operator/pull/717

PR #715, opened 2022-11-01
Fixed VF Device ID in manifests configmap for BCM57414

vrindle:


zshi-redhat:
@vrindle Is there a bug associated with this PR?

zshi-redhat:
> > @vrindle Is there a bug associated with this PR?
> 
> @zshi-redhat this is the bug [https://bugzilla.redhat.com/show_bug.cgi?id=2013986](url)

@vrindle Could you double check if there are cards with VF ID being: 16dc ? I remember this model of card may support several different VF IDs.

vrindle:
> > > @vrindle Is there a bug associated with this PR?
> > 
> > 
> > @zshi-redhat this is the bug [https://bugzilla.redhat.com/show_bug.cgi?id=2013986](url)
> 
> @vrindle Could you double check if there are cards with VF ID being: 16dc ? I remember this model of card may support several different VF IDs.

@zshi-redhat I did double check it. VF ID being 16c1 isn't right because it is the "NetXtreme-E RDMA Virtual Function". We want the "NetXtreme-E Ethernet Virtual Function" which has the correct VF ID being 16dc. Here is the link that I used: [https://cateee.net/lkddb/web-lkddb/BNXT.html](url)

wizhaoredhat:
Upstream:
https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/deployment/sriov-network-operator/templates/configmap.yaml
Has a device that is quite similar to the one you are changing but under a different name...

vrindle:
> Upstream: https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/deployment/sriov-network-operator/templates/configmap.yaml Has a device that is quite similar to the one you are changing but under a different name... Broadcom_bnxt_BCM57414_2x25G: "14e4 16d7 16dc"

Yes, I believe the upstream version is correct. This is not the correct VF device ID, we do not want the RDMA virtual function.

vrindle:
> Upstream: https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/deployment/sriov-network-operator/templates/configmap.yaml Has a device that is quite similar to the one you are changing but under a different name... Broadcom_bnxt_BCM57414_2x25G: "14e4 16d7 16dc"

I think those are the same @wizhaoredhat. https://www.broadcom.com/products/ethernet-connectivity/network-adapters/bcm57414-50g-ic says that this model is "the NetXtreme®-E Series BCM57414". I think the names are different here, but they are both referring to the same device in upstream and downstream.

wizhaoredhat:
@vrindle Should we have the names being consistent between U/S and D/S?

bn222:
The bot should not be happy, the PR needs to be named properly so that it knows which bz or jira is associated with it. Then /jira-refresh should make it happy. I'm surprised that there is no message. Can you add yourself to OWNERS?

PR #719, opened 2022-11-15
Bump config to 4.13

locriandev:
Ref. https://issues.redhat.com/browse/ART-5087

### DIFF
--- a/.ci-operator.yaml
+++ b/.ci-operator.yaml
@@ -1,4 +1,4 @@
 build_root_image:
   name: release
   namespace: openshift
-  tag: rhel-8-release-golang-1.18-openshift-4.12
+  tag: rhel-8-release-golang-1.18-openshift-4.13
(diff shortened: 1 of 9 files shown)


SchSeba:
Thanks for the PR!

/retest

.ci-operator.yaml
```diff
@@ -1,4 +1,4 @@
 build_root_image:
   name: release
   namespace: openshift
-  tag: rhel-8-release-golang-1.18-openshift-4.12
+  tag: rhel-8-release-golang-1.18-openshift-4.13
```
SchSeba, line 4: I think the image is `rhel-8-release-golang-1.19-openshift-4.13`

also please update the Dockerfiles with the rhel7 postfix

locriandev: Bumping golang is not in the scope of this PR. This is just for aligning the operator config to what ART expects. Same applies for Dockerfile changes

PR #732, opened 2022-12-07
OCPBUGS-4140: Fix machine config update in sriovnetworkpoolconfig_controller

wizhaoredhat:
These fixes only apply to Openshift clusters with the machine config operator.

The controller was not detecting changes to the machine config files. Firstly if the existing MachineConfig's ignition JSON differs from the rendered MachineConfig's ignition JSON then we should update the MachineConfig. Then we should update the MachineConfig with the rendered MachineConfig. This logic had an incorrect implementation priorly.

(cherry picked from commit 45257c4ff61285cf3c153ce3df4b7a617fa0358d)

### DIFF
--- a/controllers/sriovnetworkpoolconfig_controller.go
+++ b/controllers/sriovnetworkpoolconfig_controller.go
@@ -1,9 +1,10 @@
 package controllers
 
 import (
-	"bytes"
 	"context"
+	"encoding/json"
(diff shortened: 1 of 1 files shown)


wizhaoredhat:
@zshi-redhat Could you help with some of the labels that tide wants to see?

PR #739, opened 2022-12-14
Add script to switch BF2 between NIC and DPU mode

bn222:
This patch adds a new script called bf2-switch-mode.sh that allows the user to easily switch the BF2 between Network Interface Controller (NIC) and Data Processing Unit (DPU) modes. This is useful for optimizing the BF2 for different workloads.

Co-authored-by: Costin Gamenț <costin.gament@gmail.com>
Signed-off-by: Balazs Nemeth <bnemeth@redhat.com>

### DIFF
--- a/bindata/scripts/bf2-switch-mode.sh
+++ b/bindata/scripts/bf2-switch-mode.sh
@@ -1,0 +1,82 @@
+#!/bin/bash
+# This script can be used by any actor to switch the Bluefield
+# to NIC or DPU mode. An example to use it is by overriding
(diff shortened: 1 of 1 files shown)


SchSeba:
This is a d/s only as it was not approved in the u/s repo.
we just use the sriov-config-daemon as a container that have msfconfig binary and it's shipped by OCP already.

In the future, we are going to revert this PR and introduce a method to allow the sriov operator to change the card mode.

/lgtm

PR #742, opened 2022-12-27
OCPBUGS-5178: Fix passing device to mstconfig

bn222:


zshi-redhat:
/lgtm
downstream only fix.

EmilienM:
@bn222 e2e jobs fail for the same reason:
```
 Timed out waiting for the catalog source oo-pf7fs to become ready after 10 minutes. 
```

Sounds like unrelated to this patch, and it affects all e2e (not only OpenStack).

zshi-redhat:
Operator installation failed in the operator-e2e test, I wonder if it is due to 4.13 sriov operator catalog not ready for testing in CI. (operator-e2e test passes in the 4.12 PR).

bn222:
> Operator installation failed in the operator-e2e test, I wonder if it is due to 4.13 sriov operator catalog not ready for testing in CI. (operator-e2e test passes in the 4.12 PR).

This is what I was thinking too. I think it might make sense to bypass the bot to get this one in before 4.13 catalogue becomes available.

bn222:
We will know after GA date. If the issue disappears by then,  the we know why. we already have something in jira to track that we need to address a related issue.

PR #751, opened 2023-03-02
WIP: Support Graceful Shutdown

jerpeter1:
If the kubelet where the the sriov pod is running has gracefulShutdown configured, we'll delay in preStop for a while if and while `/tmp/sriov-delay-shutdown exists` in the host filesystem, up to a maximum wait that's less than 15 minutes (which is now specified in the daemonset pod's terminationGracePeriodSeconds).


jerpeter1:
Exploring this as a possibly less-risky alternative to involving signal handlers to supporting graceful shutdown. If a node's kubelet doesn't have gracefulShutdown configured, this should do nothing, and if it does, it should delay preStop for a little while.

SchSeba:
Hi @jerpeter1 I am closing this one we continue to work on the u/s and then we do a full sync between all the PRs

/close

PR #758, opened 2023-03-16
Fix bundle & manifest generation for vDPA CRDs

lmilleri:
See https://issues.redhat.com/browse/OCPBUGS-10381


lmilleri:
@bn222 @zshi-redhat please add the ok-to-test label

PR #778, opened 2023-05-15
Fix issue with BF2 mode switch

wizhaoredhat:
The BF2 mode switch script will always return a code used for signaling reboots even if the BF2 is already in the expected mode. The changes needed is to only return the 120 code when a reboot is needed.

Co-authored-by: Salvatore Daniele <sdaniele@redhat.com>
Signed-off-by: William Zhao <wizhao@redhat.com>

### DIFF
--- a/bindata/scripts/bf2-switch-mode.sh
+++ b/bindata/scripts/bf2-switch-mode.sh
@@ -7,30 +7,44 @@
 # --privileged --volume /dev:/dev --entrypoint \
 # /bindata/scripts/bf2-switch-mode.sh http://url/of/image && \
 # systemctl reboot
+#
+# Returns 0   if no reboot is necessary however devices may not be found.
+# Returns 120 if reboot is necessary after applying firmware changes.
 
-set -e
+set -o pipefail
 
 mode=${1:-query}
 device=${2:-"$(mstconfig q | grep "Device:" | awk 'BEGIN {FS="/";} { print $6 }')"}
+if [ $? -ne 0 ]; then
+    echo "Failed to query for device."
+    exit 0
+fi
 
 if [ -z "$device" ]; then
     echo "Can't find device."
-    exit 120
+    exit 0
 else
     echo Found device: "${device}"
 fi
 
 type=$(mstconfig -d $device q | grep "Device type:" | awk '{ print $3}')
+if [ $? -ne 0 ]; then
+    echo "Failed to query ${device} for device type."
+    exit 0
+fi
 
 if [ "${type}" != "BlueField2" ]; then
     echo "Device is not a Bluefield2"
-    exit 120
+    exit 0
 fi
 
 fwreset () {
     mstfwreset -y -d "${device}" reset
+    if [ $? -ne 0 ]; then
+        echo "Failed to run mstfwreset for ${device}."
+    fi
     echo "Switched to ${mode} mode."
-    exit 0
+    exit 120
 }
 
 query () {
@@ -40,6 +54,10 @@ query () {
         INTERNAL_CPU_ESWITCH_MANAGER \
         INTERNAL_CPU_IB_VPORT0 \
         INTERNAL_CPU_OFFLOAD_ENGINE | tail -n 5 | awk '{print $2}' | xargs echo)
+    if [ $? -ne 0 ]; then
+        echo "Failed to get the current config for ${device}."
+        exit 0
+    fi
     echo "Current DPU configuration: ${current_config}"
 }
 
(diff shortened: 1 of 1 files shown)


bn222:
This isn't right. We should only return 0 if a switch occurred. The way the docs talks about this is as follows:

```
apiVersion: machineconfiguration.openshift.io/v1
kind: MachineConfig
metadata:
  labels:
    machineconfiguration.openshift.io/role: sriov
  name: 99-bf2-dpu
spec:
  config:
    ignition:
      version: 3.2.0
    storage:
      files:
      - contents:
          source: data:text/plain;charset=utf-8;base64,ZmluZF9jb250YWluZXIoKSB7CiAgY3JpY3RsIHBzIC1vIGpzb24gfCBqcSAtciAnLmNvbnRhaW5lcnNbXSB8IHNlbGVjdCgubWV0YWRhdGEubmFtZT09InNyaW92LW5ldHdvcmstY29uZmlnLWRhZW1vbiIpIHwgLmlkJwp9CnVudGlsIG91dHB1dD0kKGZpbmRfY29udGFpbmVyKTsgW1sgLW4gIiRvdXRwdXQiIF1dOyBkbwogIGVjaG8gIndhaXRpbmcgZm9yIGNvbnRhaW5lciB0byBjb21lIHVwIgogIHNsZWVwIDE7CmRvbmUKISBzdWRvIGNyaWN0bCBleGVjICRvdXRwdXQgL2JpbmRhdGEvc2NyaXB0cy9iZjItc3dpdGNoLW1vZGUuc2ggIiRAIgo=
        mode: 0755
        overwrite: true
        path: /etc/default/switch_in_sriov_config_daemon.sh
    systemd:
      units:
      - name: dpu-switch.service
        enabled: true
        contents: |
          [Unit]
          Description=Switch BlueField2 card to NIC/DPU mode
          RequiresMountsFor=%t/containers
          Wants=network.target
          After=network-online.target kubelet.service
          [Service]
          SuccessExitStatus=0 120
          RemainAfterExit=True
          ExecStart=/bin/bash -c '/etc/default/switch_in_sriov_config_daemon.sh nic || shutdown -r now' 
          Type=oneshot
          [Install]
          WantedBy=multi-user.target
```

The `ExecStart` line is important for our discussion. As you can see, it will reboot the node depending on the return value. If you return 0 in the case you mention, the node will get into a boot-loop.

bn222:
Don't you have a boot loop in that case today then?

SalDaniele:
That's what we expected but for some reason this is not the case. I don't see this ExecStart step run again in the logs afterwards. The nodes do seem to reboot an additional time after the CDA installation completes however, which is why I noticed this issue.

bn222:
You're patch still doesn't make sense to me. fwreset already has exit 0, so the two execution paths are not differentiated anymore. A reboot will occur when you return 0. Can you explain in more detail how this change should work?

SalDaniele:
My understanding was that the behavior of `ExecStart=/bin/bash -c '/etc/default/switch_in_sriov_config_daemon.sh nic || shutdown -r now'` will be that in the event switch_in_sriov_config_daemon.sh is successful, a reboot will not occur (a successful exit status of 0 will mean that "shutdown -r now" will not be called, such as in the case of fwreset()).

Currently, if we call bf2_switch_mode to switch a BF already in NIC mode into NIC mode, we will exit 120, which will trigger a reboot.

This change should modify this behavior so that a reboot will not occur if we try to switch a NIC mode bf into NIC mode IIUC (@wizhaoredhat perhaps I am misunderstanding your thinking?).

Is there some reason I am missing that we would want to reboot in that scenario?

bn222:
Looking at the change, it will cause both the branch where we have

`Already in NIC mode.`
and
`Switching to NIC mode.`

to return 0.


In one of those two, we need to have a reboot, but we don't distinguish anymore. Did you test your patch on a fresh installation?

bindata/scripts/bf2-switch-mode.sh
```diff
@@ -70,6 +91,9 @@ case "${mode}" in
                 INTERNAL_CPU_ESWITCH_MANAGER=EXT_HOST_PF \
                 INTERNAL_CPU_IB_VPORT0=EXT_HOST_PF \
                 INTERNAL_CPU_OFFLOAD_ENGINE=DISABLED
+            if [ $? -ne 0 ]; then
+                echo "Failed to config for NIC mode on ${device}."
```
bn222, line 95: Should return 0 in this case. Same for DPU mode.

wizhaoredhat:
Although confusing the systemd script negates the return value from this script. Such that any error code that is non-zero will not reboot the BF2. Any error code that returns 0 will reboot the BF2.

PR #780, opened 2023-05-22
Sync 2023-05-19

liornoy:
This PR does a full sync from 2023-03-23 to 2023-05-19


SchSeba:
we leave the [Intel x550 to supported NICs](https://github.com/openshift/sriov-network-operator/pull/780/commits/95cfca53751fdf5816b7d4c39836c718d1297e7d) out of the bundle as there is no request to support this nic on OCP

PR #781, opened 2023-05-31
OCPBUGS-14154: Add pciutils dependency

SalDaniele:
Without pciutils, we will get the following error message when running mstfwreset:

Continue with reset?[y/N] y
Failed
-E- failed to run 'setpci -s 0000:c9:02.0 0x0.w'.

The setpci command is part of the pciutils package.


wizhaoredhat:
@SchSeba PTAL.

For DPU mode switch to NIC mode, we need to use mstfwreset otherwise the switch does not occur and the switch to NIC mode would silently fail.

wizhaoredhat:
@bn222 PTAL

PR #784, opened 2023-06-01
[release-4.10] OCPBUGS-14360: Continue node drain after reboot

wizhaoredhat:
Backport the following commits:
- 'Continue node drain after reboot' d48291194e861bcbcb575b9884d6a0a7a615d461
- 'Annotate node only if there is no existing annotation' 008d2fe669e8922f783b43c490d7a5820bc55daa

After a node reboot, the number of VFs would be zero, then `reqDrain` would be set and we will need to drain also after a reboot. This behavior is caused by this
commit backport  https://github.com/openshift/sriov-network-operator/commit/7d099d32f3a38e89a56f73807ce60aa26b35cef4 . This will cause the drain lock to be attempted to be taken again. However a deadlock would occur if another node already took the lock at the same time. The  'Continue node drain after reboot' commit solves this issue by checking if we have the drain node annotation which means we are safe to proceed draining without taking the lock.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -503,20 +503,23 @@ func (dn *Daemon) nodeStateSyncHandler(generation int64) error {
 	}
 
 	if reqDrain {
-		if !dn.disableDrain {
-			ctx, cancel := context.WithCancel(context.TODO())
-			defer cancel()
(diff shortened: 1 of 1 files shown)


wizhaoredhat:
/cc @zhaozhanqi
/cc @SchSeba 
PTAL

PR #786, opened 2023-06-05
OCPBUGS-14591: Allow for multiple devices

SalDaniele:
Currently if multiple devices are detected, bf2-switch-mode will fail, and provide the misleading error "Can't find bluefield device".

Instead, check each device to verify if it is a Bluefield2, and if so, apply new configurations.

### DIFF
--- a/bindata/scripts/bf2-switch-mode.sh
+++ b/bindata/scripts/bf2-switch-mode.sh
@@ -11,26 +11,32 @@
 set -e
 
 mode=${1:-query}
-device=${2:-"$(mstconfig q | grep "Device:" | awk 'BEGIN {FS="/";} { print $6 }')"}
+devices=${2:-"$(mstconfig q | grep "Device:" | awk 'BEGIN {FS="/";} { print $6 }')"}
 
-if [ -z "$device" ]; then
-    echo "Can't find device."
+for device in $devices; do
+    echo "Found device: $device"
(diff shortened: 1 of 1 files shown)


SalDaniele:
@bn222 @wizhaoredhat PTAL

wizhaoredhat:
LGTM, thanks Sal!

bindata/scripts/bf2-switch-mode.sh
```diff
@@ -11,26 +11,32 @@
...
-    exit 120
-else
-    echo Found device: "${device}"
-fi
+for device in $devices; do
+    echo "Found device: $device"
+    type=$(mstconfig -d $device q | grep "Device type:" | awk '{ print $3}')
 
-type=$(mstconfig -d $device q | grep "Device type:" | awk '{ print $3}')
+    if [ "${type}" != "BlueField2" ]; then
```
wizhaoredhat, line 20: In anticipation for the BF3. Should we just do a regex here?

bn222: Let's not anticipate too much. Just keep it hard-coded for Bluefield 2 for now.

SalDaniele: Sounds good. Is it possible to get lgtm/approved on this PR? Tested and verified it works on a cluster with multiple nics

PR #788, opened 2023-06-06
[release-4.10] OCPBUGS-14359: Skip the Redhat virtual nic from udev rule

wizhaoredhat:
we need to remove the Red Hat Virtio network device from the udev rule configuration if we don't remove it when running the config-daemon on a virtual node it will disconnect the node after a reboot even that the operator should not be installed on virtual environments that are not openstack we should not destroy the cluster if the operator is installed there

(cherry picked from commit fe32cadb9580ae3e205cdf63f8905c31f669b253) 
Conflicts:
   There was refactoring done in the commit d20be75f5b92e2b4db155469d324cf837dd41ae2
   under the PR 688 which caused some conflicts. However we should just
   use the variable name "new_content" for this small backport.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -1073,7 +1073,18 @@ func tryCreateNMUdevRule() error {
 	dirPath := "/host/etc/udev/rules.d/"
 	filePath := dirPath + "10-nm-unmanaged.rules"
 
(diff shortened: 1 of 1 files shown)


wizhaoredhat:
For reference: https://github.com/openshift/sriov-network-operator/pull/688/commits/d20be75f5b92e2b4db155469d324cf837dd41ae2

PR #789, opened 2023-06-09
OCPBUGS-13724: Update version to 4.14

vrindle:


wizhaoredhat:
/cc @SchSeba 
/cc @bn222 
PTAL.

LGTM

bn222:
can you check if we've missed anything by running `make manifests` and `make bundle`?

vrindle:
> can you check if we've missed anything by running `make manifests` and `make bundle`?

I ran it and it should be good.

PR #791, opened 2023-06-14
OCPBUGS-14954: Fix multiple dev bug 4.13

SalDaniele:
Currently if multiple devices are detected, bf2-switch-mode will fail, and provide the misleading error "Can't find bluefield device".

Instead, check each device to verify if it is a Bluefield2, and if so, apply new configurations.


SalDaniele:
@zhaozhanqi Backport of https://github.com/openshift/sriov-network-operator/pull/786 PTAL

PR #792, opened 2023-06-15
Modifying config-daemon to exit if platform type cannot be determined.

bmorettodama:
With the current configuration, the config-daemon always defaults to BareMetal if it fails to obtain the node object. We should instead try again, and if we fail to determine the platform, then we should exit. Defaulting to BareMetal will cause issues if the platform is GKE or OpenShift, and we should be using the virtual plugin.


SchSeba:
Hi @bmorettodama all the PRs should go to the upstream repo first for review by the community.

PR #808, opened 2023-07-28
OCPBUGS-17047: Add support for BCM57504

SalDaniele:
Enable legacy SR-IOV in Broadcom BCM57504 NetXtreme-E device.

PR #809, opened 2023-08-02
Partially revert BCM57504 update

SalDaniele:
https://github.com/openshift/sriov-network-operator/pull/808 added support for Broadcom BCM57504. Since no vendor from Broadcom is currently participating in maintaining support for this card upstream, this change has been made downstream only.

In order to avoid future merge conflicts with u/s, this change should be added in d/s only files. This commit reverts any changes to u/s config maps.

### DIFF
--- a/bundle/manifests/supported-nic-ids_v1_configmap.yaml
+++ b/bundle/manifests/supported-nic-ids_v1_configmap.yaml
@@ -1,7 +1,6 @@
 apiVersion: v1
 data:
   Broadcom_bnxt_en_BCM57414_NetXtreme-E: 14e4 16d7 16c1
-  Broadcom_bnxt_en_BCM57504_NetXtreme-E: 14e4 1751 1806
   Broadcom_bnxt_en_BCM57508_NetXtreme-E: 14e4 1750 1806
   Intel_i40e_10G_X710_BACKPLANE: 8086 1581 154c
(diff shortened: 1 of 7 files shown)


bundle/manifests/supported-nic-ids_v1_configmap.yaml
```diff
@@ -1,7 +1,6 @@
 apiVersion: v1
 data:
   Broadcom_bnxt_en_BCM57414_NetXtreme-E: 14e4 16d7 16c1
-  Broadcom_bnxt_en_BCM57504_NetXtreme-E: 14e4 1751 1806
```
SchSeba, line 4: this one you can add.

this file only exist on d/s

SchSeba (commented):
just a small comment

bundle/manifests/supported-nic-ids_v1_configmap.yaml
SalDaniele: Rhis file is autogenerated using config/manifests/bases/sriov-network-operator_configmap.yaml, so it will not be present after running "make bundle manifests"

SchSeba: you are right sorry!

SchSeba (approved):
/lgtm
/approve

PR #811, opened 2023-08-11
OCPBUGS-14499: Sync 20230811

zeeke:
clean merge


EmilienM:
please ignore the openstack jobs for now, we're investigating why they're red on 4.14. Not related to SRIOV network operator.
fyi @SchSeba too

PR #815, opened 2023-08-24
OCPBUGS-16906: Never reduce MTU on PF

bn222:
In case a policy specifying a different MTU, in the past, the MTU of the PF associated with the VFs would also be updated. Since reducing the MTU on PF could cause issues when the PF is used by other things too (i.e. primary network), reducing the MTU could cause connectivity issues. Therefore, take the conservative approach of never reducing the MTU on the PF when configuring the MTU on the VF.

### DIFF
--- a/cmd/webhook/start.go
+++ b/cmd/webhook/start.go
@@ -4,7 +4,7 @@ import (
 	"crypto/tls"
 	"encoding/json"
 	"fmt"
-	"io/ioutil"
+	"io"
 	"net/http"
 
 	"github.com/fsnotify/fsnotify"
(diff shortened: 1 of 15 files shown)


bn222:
@zeeke ptal

zeeke:
Hi @bn222 , can you do a merge sync instead of cherry-picking?

If you want to import the minimum amount of commits for OCPBUGS-16909, you can try with:

```
git merge 4dd75bc7ae339e8397266580ff1b97a6b2125bdb
```

That way, we guarantee the commit history is the same as the upstream repository and the next sync won't cause (at least has less chance) conflicts.

PR #816, opened 2023-08-31
Revert "Updating images to pull from github instead of quay.io"

SalDaniele:
This reverts commit 3eb2f83d98d091e48ba2ff3a8f83d86c5da004c6.

We should not use upstream images downstream as ds only resources will be missing when deploying with make deploy-setup. (i.e. /bindata/scripts/bf2-switch-mode.sh)


SchSeba:
/lgtm
/approve

d/s only change

PR #818, opened 2023-09-04
OCPBUGS-18501: SR-IOV should reject bad MTU config in policy

bn222:
In case a policy specifying a different MTU, in the past, the MTU of the PF associated with the VFs would also be updated. Since reducing the MTU on PF could cause issues when the PF is used by other things too (i.e. primary network), reducing the MTU could cause connectivity issues. Therefore, take the conservative approach of never reducing the MTU on the PF when configuring the MTU on the VF.

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -299,7 +299,7 @@ func configSriovDevice(iface *sriovnetworkv1.Interface, ifaceStatus *sriovnetwor
 		}
 	}
 	// set PF mtu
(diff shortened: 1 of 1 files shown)


SchSeba:
Hi @bn222 question did you did a cherry-pick from the master one?
I was expecting a cherry-pick to maintain the same commit hash

SchSeba:
ok I just validate all good

/lgtm
/approve

PR #820, opened 2023-09-05
Downstream Sync 2023-09-05

liornoy:
This PR syncs with d/s mainly to bring the K8S bump, that is urgent for
resolving a dependency conflict in the cnf-tests.


zeeke:
Please, use the `merge` sync method instead of `cherry-pick`.

We recently switched to the merge method as:
- It's less error-prone, there is almost no way of forgetting a commit
- It makes downstream commit SHAs to be the same as upstream

liornoy:
@zeeke done

zeeke:
Thanks @liornoy 

/lgtm
/approve

PR #821, opened 2023-09-11
Sync 20230911

bn222:


Makefile
```diff
@@ -177,23 +172,6 @@ skopeo:
 fakechroot:
 	if ! which fakechroot; then if [ -f /etc/redhat-release ]; then dnf -y install fakechroot; elif [ -f /etc/lsb-release ]; then sudo apt-get -y update; sudo apt-get -y install fakechroot; fi; fi
 
-# Generate bundle manifests and metadata, then validate generated files.
-.PHONY: bundle
-bundle: manifests
```
zeeke, line 182: Will this rule be rewritten in a downstream-only Makefile? Do we have to change some openshift/release CI configuration to avoid having broken builds?

bn222: Added `Makefile.bundle`. 

@joepvd , do we need to change something in release/ci for this to work?

PR #837, opened 2023-10-02
OCPBUGS-19908: virtual: get real PCI address for each device found

EmilienM:
We can't rely on the PCI address from the metadata so we will lookup the real PCI address
for the NIC that matches the MAC address.

Libvirt/QEMU cannot guarantee that the address specified in the XML will match the address seen by the guest.
This is a well known limitation: https://libvirt.org/pci-addresses.html
When using the q35 machine type, it highlights this issue due to the change from using PCI to PCI-E bus for virtual devices.

With that said, the PCI value in Nova Metadata is a best effort hint due to the limitations mentioned above. Therefore
we will lookup the real PCI address for the NIC that matches the MAC address.

This PR also adds  Virtio 1.0 network device as supported

https://devicehunt.com/view/type/pci/vendor/1AF4/device/1041
https://qemu.readthedocs.io/en/master/specs/pci-ids.html

Modern QEMU brings up virtio 1.0 network device, we need to support that
for DPDK.

### DIFF
--- a/pkg/utils/utils_virtual.go
+++ b/pkg/utils/utils_virtual.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"path/filepath"
 	"strconv"
+	"strings"
 
 	"github.com/golang/glog"
 	"github.com/hashicorp/go-retryablehttp"
@@ -118,6 +119,27 @@ func GetOpenstackData(useHostPath bool) (metaData *OSPMetaData, networkData *OSP
 	if err != nil {
 		metaData, networkData, err = getOpenstackDataFromMetadataService()
 	}
+
+	// We can't rely on the PCI address from the metadata so we will lookup the real PCI address
+	// for the NIC that matches the MAC address.
+	//
+	// Libvirt/QEMU cannot guarantee that the address specified in the XML will match the address seen by the guest.
+	// This is a well known limitation: https://libvirt.org/pci-addresses.html
+	// When using the q35 machine type, it highlights this issue due to the change from using PCI to PCI-E bus for virtual devices.
+	//
+	// With that said, the PCI value in Nova Metadata is a best effort hint due to the limitations mentioned above. Therefore
+	// we will lookup the real PCI address for the NIC that matches the MAC address.
+	for i, device := range metaData.Devices {
+		realPCIAddr, err := getPCIAddressFromMACAddress(device.Mac)
+		if err != nil {
+			return metaData, networkData, fmt.Errorf("GetOpenstackData(): error getting PCI address for device %s: %w", device.Mac, err)
+		}
+		if realPCIAddr != device.Address {
+			glog.V(2).Infof("GetOpenstackData(): PCI address for device %s does not match Nova metadata value %s, it'll be overwritten with %s", device.Mac, device.Address, realPCIAddr)
+		}
+		metaData.Devices[i].Address = realPCIAddr
+	}
+
 	return metaData, networkData, err
 }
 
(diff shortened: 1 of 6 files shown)


EmilienM:
/hold

under testing

MaysaMacedo:
Nice catch!
Should this fix also be made available upstream?
https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/pkg/utils/utils_virtual.go#L116-L122

EmilienM:
> Nice catch! Should this fix also be made available upstream? https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/pkg/utils/utils_virtual.go#L116-L122

https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/516

This PR is just for internal testing for now.

pkg/utils/utils_virtual.go
```diff
@@ -205,6 +226,33 @@ func getOpenstackDataFromMetadataService() (metaData *OSPMetaData, networkData *
...
+// getPCIAddressFromMACAddress returns the PCI address of a device given its MAC address
+func getPCIAddressFromMACAddress(macAddress string) (string, error) {
+	nics, err := ghw.Network()
+	if err != nil {
+		return "", fmt.Errorf("error getting network info: %v", err)
+	}
+
+	var pciAddress string
+	for _, nic := range nics.NICs {
+		if nic.MacAddress == macAddress {
```
mandre, line 238: I don't know how these are returned, but wouldn't it be safer to lowercase the MAC addresses (both `nic.MacAddress` and `macAddress`) before comparing?

pkg/utils/utils_virtual.go
```diff
@@ -118,6 +118,27 @@ func GetOpenstackData(useHostPath bool) (metaData *OSPMetaData, networkData *OSP
...
+	// Libvirt/QEMU cannot guarantee that the address specified in the XML will match the address seen by the guest.
+	// This is a well known limitation: https://libvirt.org/pci-addresses.html
+	// When using the q35 machine type, it highlights this issue due to the change from using PCI to PCI-E bus for virtual devices.
+	//
+	// With that said, the PCI value in Nova Metadata is a best effort hint due to the limitations mentioned above. Therefore
+	// we will lookup the real PCI address for the NIC that matches the MAC address.
+	for _, device := range metaData.Devices {
+		realPCIAddr, err := getPCIAddressFromMACAddress(device.Mac)
+		if err != nil {
+			return metaData, networkData, fmt.Errorf("GetOpenstackData(): error getting PCI address for device %s: %v", device.Mac, err)
```
mandre, line 134: I believe it's more idiomatic to use the `%w` verb for errors.

pkg/utils/utils_virtual.go
```diff
@@ -118,6 +118,27 @@ func GetOpenstackData(useHostPath bool) (metaData *OSPMetaData, networkData *OSP
...
+		}
+		if realPCIAddr != device.Address {
+			glog.Warningf("GetOpenstackData(): PCI address for device %s does not match Nova metadata value %s, it'll be overwritten with %s", device.Mac, device.Address, realPCIAddr)
```
mandre, line 137: `Warning` level seems too much, as we can't rely on the nova metadata for the PCI address and this would then be an expected scenario to have the PCI address differ. What do you think of lowering the log level?

mandre (commented):
Yes! We'll need to backport that too.

EmilienM:
@mandre thanks for your review, I'll address your comments shortly. I didn't have the time today due to testing this PR in real environment.

> We'll need to backport that too.
:+1:

EmilienM:
Weird, the webhook is failing: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_sriov-network-operator/837/pull-ci-openshift-sriov-network-operator-master-e2e-openstack-nfv/1709659020530290688/artifacts/e2e-openstack-nfv/gather-extra/artifacts/pods/openshift-sriov-network-operator_operator-webhook-7fvmg_webhook-server.log

EmilienM:
taken care here: https://github.com/openshift/sriov-network-operator/pull/842

PR #842, opened 2023-10-23
OCPBUGS-19908: Downstream sync 2023-10-23

zeeke:
clean merge from upstream without conflicts

ran `make -f Makefile.bundle generate bundle` to align downstream manifests for
- `externallyManaged` field in sriovnetworknodestate and sriovnetworknodepolicy (cc @SchSeba)
- `vlanProto` field in sriovnetwork (cc @mlguerrero12)

cc @EmilienM


EmilienM:
out of curiosity but not blocking it:
/test e2e-openstack-nfv

EmilienM:
@zeeke could we have a sync on 4.14 also, thanks.

zeeke:
> @zeeke could we have a sync on 4.14 also, thanks.

There is no sync for older branches. We can cherry-pick specific commits and create a pull request targeting release-4.14

PR #849, opened 2023-10-24
OCPBUGS-21192: disable HTTP2 by default

cgoncalves:


cgoncalves:
@zhaozhanqi please review.

SchSeba:
Hi @cgoncalves there is a problem with one of the libs require a newer go version.

btw I think 4.11 is end of life no?

cgoncalves:
@SchSeba yeah. This is an issue that @zeeke fixed upstream: https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/551. The next step would be to bring it downstream and to all supported releases.

OCP 4.11 is under maintenance support until February 10, 2024.

cgoncalves:
/close

4.11 CI is broken due to a known issue fixed in newer releases but it won't be backported. 4.11 is very soon at the end of life (February 10, 2024).

PR #867, opened 2023-11-22
OCPBUGS-22807: Hash contents of cm instead of using resource version

bn222:


bn222:
Not a clean backport. Needed to make sure that `r.syncSriovNetworkNodeState` was called correctly. The change is trivial.

wizhaoredhat:
> Not a clean backport. Needed to make sure that `r.syncSriovNetworkNodeState` was called correctly. The change is trivial.

I see it, it is that the function is missing ctx in `r.syncSriovNetworkNodeState(ctx, np, npl, ns, &node, utils.HashConfigMap(found));`

PR #873, opened 2023-12-13
OCPBUGS-24231: Fix config daemon annotation override

cgoncalves:
A downstream merge [1][2] overrode the annotations in the config daemonset. This resulted in he sriov-network-config-daemon pod missing the `target.workload.openshift.io/management` annotation which led to the pod not being pinned to the management cores when workload partitioning is enabled.

This is a downstream-only patch.

[1] https://github.com/openshift/sriov-network-operator/pull/868
[2] commit 0f177536b2db307af95d251548bd5c055e2b543c

### DIFF
--- a/bindata/manifests/daemon/daemonset.yaml
+++ b/bindata/manifests/daemon/daemonset.yaml
@@ -16,8 +16,6 @@ spec:
       maxUnavailable: 33%
   template:
     metadata:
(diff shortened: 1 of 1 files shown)


zeeke:
/lgtm

@SchSeba , @wizhaoredhat please take a look

PR #902, opened 2024-03-29
Change format of features.operators.openshift.io/*

bn222:


zshi-redhat:
/retitle OCPBUGS-31519: Change format of features.operators.openshift.io/*

PR #927, opened 2024-04-19
hack: let "depoly-setup.sh" and "undeploy.sh" fail on error

thom311:
Ignoring errors seems not best. It can make it harder to find problems with the script.

If a command in the script is known to fail for valid reasons, then that should be treated specially (for example, with "|| :"). By default, propagate any failure to the caller.

Fixes: 88441502023d ('Add RBAC for cluster resource')

---

WARNING: I did not carefully confirm, that the script does not fail commands in valid cases. If the script is know that certain commands fail for valid reasons, then this change would make the script fail. Such places should be identified, and fixed with a `|| :`.

### DIFF
--- a/hack/deploy-setup.sh
+++ b/hack/deploy-setup.sh
@@ -3,7 +3,7 @@
 # to deploy.  It assumes it is capable of login as a
 # user who has the cluster-admin role
 
-# set -euxo pipefail
+set -eo pipefail
 
 source "$(dirname $0)/common"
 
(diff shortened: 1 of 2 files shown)


zeeke:
Hi @thom311 , can you propose this fix to the upstream repository?

I'm 100% in favor of this improvement, but there is no need to have it in the openshift version only. Changes from the upstream repo are pulled and merged here regularly.

thom311:
> Hi @thom311 , can you propose this fix to the upstream repository? https://github.com/k8snetworkplumbingwg/sriov-network-operator

I didn't realize (that there is an upstream).

Will do. Thanks.

Closing.

PR #942, opened 2024-05-30
d/s: remove `spec.install.spec.deployments` from base CSV

zeeke:
File `config/manager/manager.yaml` defines the operator Deployment that is reported by the final ClusterServiceVersion (CSV) and overwrites whatever is written in the base CSV definition (`config/manifests/bases/sriov-network-operator.clusterserviceversion.yaml`). Remove the definition that is not used to make the operator's bundle build process clearer.


zeeke:
@bn222 @SchSeba please take a look

PR #974, opened 2024-07-29
OCPBUGS-37668: [release-4.15] operator: LeaderElectionReleaseOnCancel

zeeke:
backport of:
- https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/556

This ensure the lease is released by the operator as soon as it gracefully stops.


deploy/operator.yaml
```diff
@@ -41,6 +41,8 @@ spec:
           image: $SRIOV_NETWORK_OPERATOR_IMAGE
           command:
           - sriov-network-operator
+          args:
+          - --leader-elect=$OPERATOR_LEADER_ELECTION_ENABLE
```
SchSeba, line 45: do we need this in the bundle also?

zeeke: Openshift bundle already has `--leader-elect` option set by default

https://github.com/openshift/sriov-network-operator/blob/master/manifests/stable/sriov-network-operator.clusterserviceversion.yaml#L366

SchSeba: great thanks!

btw I am almost sure I already asked that :P

zeeke:
/hold

wait for the fix in:
- https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/760

PR #1018, opened 2024-10-11
[Downstream] Add new DPU/Accelerator NICs

wizhaoredhat:
We are adding the Intel NetSec Accelerator and Marvell Octeon 10 DPU


wizhaoredhat:
@bn222 @SchSeba  PTAL at these DPU/Accelerator NICs that we are supporting.

bn222:
use merge bot instead? ie. just wait?

bn222:
Or is this downstream only?

wizhaoredhat:
> Or is this downstream only?

Downstream only.

PR #1031, opened 2024-11-13
OCPBUGS-44515: relax Operator node affinity for Hypershift

EmilienM:
In the context of Hypershift (Hosted Clusters), where a Nodepool
(terminology for a worker Node in HCP) is not a control-plane or
a master Node but a worker, we can't force the

The proposal here is to relax the rule and use
`preferredDuringSchedulingIgnoredDuringExecution` instead so the scheduler
will try to find a master node or fallback on other nodes if not found.

### DIFF
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -20,7 +20,12 @@ spec:
     spec:
       affinity:
         nodeAffinity:
-          requiredDuringSchedulingIgnoredDuringExecution:
+          # In the context of Hypershift, the SR-IOV network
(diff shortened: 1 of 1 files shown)


EmilienM:
I'm really not sure it's worth keeping that affinity rule at all, but I might have missed something.

EmilienM:
Upstream PR: https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/806

EmilienM:
Will be taken care of by the bot since the upstream PR just merged
/close

PR #1052, opened 2025-01-13
[th/deploy-detect-defaults] fix `make deploy-setup` to pass and autodetect some URLs

thom311:
Currently, `make deploy-setup` fails unless we set `$SKIP_VAR_SET`.

Quote commit message:

---

hack/env.sh: fix detecting default values for images

Note that most of quay.io repositories (not all of them) are depublished
and the detection no longer works:

    FATA[0001] Error parsing image name "docker://quay.io/openshift/origin-sriov-cni": reading manifest latest in quay.io/openshift/origin-sriov-cni: unauthorized: access to the requested resource is not authorized

Together that with [1],  the calling shell scripts now `set -e`, which
exits the script in such errors.

This means, `make deploy-setup` fails unless $SKIP_VAR_SET is set. That
is even if the caller *did* set the missing environment variables for
the images. Note that doing that is not useless, because while several
quay.io repositories no longer exist, others like [2] still do.  So you
can set the ones we must, don't set $SKIP_VAR_SET, and let the others be
detected as in the past.

Fix this by reworking the detection.

The patch does not change URLs like "quay.io/openshift/origin-sriov-cni"
because there is no better default. If the caller doesn't set the
environment variables like SRIOV_CNI_IMAGE, then deploy is going to fail
either way.

[1] 624a16feb327 ('Let "depoly-setup.sh" and "undeploy.sh" fail on error')
[2] https://quay.io/openshift/origin-kube-rbac-proxy

### DIFF
--- a/hack/env.sh
+++ b/hack/env.sh
@@ -1,35 +1,48 @@
 #!/bin/bash
 
+_url_detect() {
+    local varname="$1"
+    local url="$2"
+    local digest
+    local value=
+
+    if [ "$EMPTY_IS_VALID" = 1 ] ; then
+        [ -n "${!varname+set}" ] && return 0
+    else
+        [ -n "${!varname}" ] && return 0
+    fi
+
+    if [ -n "$url" ] ; then
+        digest="$(skopeo inspect "docker://$url" | jq --raw-output '.Digest')" || :
+        if [ -z "$digest" ] ; then
+            echo "Failure to detect \$$varname at $url. Set the variable"
+            return 1
+        fi
+        value="$url@$digest"
+    fi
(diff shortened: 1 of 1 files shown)


SchSeba:
Hi @thom311,
we have this changes in u/s?

thom311:
> we have this changes in u/s?

Yes, the commit 1-of-4 " hack/env.sh: move checking of environment variables outside SKIP_VAR_SET block" is from upstream. We should first re-import upstream, then I rebase this PR and the commit will be dropped.

thom311:
@SchSeba thank you!

PR #1065, opened 2025-03-18
SRIOV PF got unbind instead of VF in case of IB link type

wizhaoredhat:
This PR back-ports the PR: https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/797

Which fixes a discrepancy where in between versions 4.16 and 4.17, that the unbinding of the VF chooses the PF instead causing failures.


wizhaoredhat:
@SchSeba @zeeke  PTAL

PR #1089, opened 2025-05-13
CNF-17884 Operator's network policies

zeeke:
Operator's network policies added to the bundle


mlguerrero12:
it seems it is not needed, indeed

PR #1093, opened 2025-05-27
Add Intel E825c Network Adapaters

m-naught:


bn222:
added upstream first?

bn222:
lgtm but we have a bot that syncs with upstream and we need to be careful not to break that

bn222:
/unhold


@m-naught needs rebase

PR #1104, opened 2025-06-24
ds: bump operator to 4.20

zeeke:
and generate manifest bundle


zeeke:
/retitle OCPBUGS-57973: ds: bump operator to 4.20

PR #1195, opened 2026-04-22
OCPBUGS-80867,OCPBUGS-80866,OCPBUGS-80863: Bump google.golang.org/grpc to v1.79.3 to fix CVE-2026-33186

kunalmemane:
- Bump google.golang.org/grpc to v1.79.3 to fix CVE-2026-33186
- Upgrade go directive to 1.24 to align with [.ci-operator.yaml](https://github.com/openshift/sriov-network-operator/blob/release-4.21/.ci-operator.yaml)
- Fix go vet printf analyzer errors (fmt.Errorf(variable) with no format args) that break test-pkg in CI after the `go 1.23` to `go 1.24` directive bump.
- Upgrade controller-gen to v0.18.0 in Makefile


pkg/host/internal/kernel/kernel.go
```diff
@@ -60,7 +60,7 @@ func (k *kernel) IsKernelModuleLoaded(kernelModuleName string) (bool, error) {
 	log.Log.V(2).Info("IsKernelModuleLoaded():", "stdout", stdout)
 	if len(stderr) != 0 {
 		log.Log.Error(err, "IsKernelModuleLoaded(): failed to check if kernel module is loaded", "name", kernelModuleName, "stderr", stderr)
-		return false, fmt.Errorf(stderr)
+		return false, fmt.Errorf("%s", stderr)
```
SchSeba: why we need this changes?
are they the same u/s and on main? if not please open a small u/s PR and we will merge it before.

I prefer to not have code difference as much possible and the changes are small and not related only to a backport

pkg/host/internal/sriov/sriov.go
```diff
@@ -412,15 +412,15 @@ func (s *sriov) checkExternallyManagedPF(iface *sriovnetworkv1.Interface) error
 			"functions %d but the policy is configured as ExternallyManaged for device %s",
 			iface.NumVfs, currentNumVfs, iface.PciAddress)
 		log.Log.Error(nil, errMsg)
-		return fmt.Errorf(errMsg)
+		return errors.New(errMsg)
```
SchSeba, line 415: same as the other comment

go.mod
```diff
@@ -1,8 +1,8 @@
 module github.com/k8snetworkplumbingwg/sriov-network-operator
 
-go 1.23.0
+go 1.24.0
 
-toolchain go1.23.4
+toolchain go1.26.2
```
SchSeba: do we really want to bump the toolchain to a newer major go version?
is it possible to avoid this?

Makefile
```diff
@@ -145,7 +145,7 @@ mock-generate: gomock
 
 CONTROLLER_GEN = $(BIN_DIR)/controller-gen
 controller-gen: ## Download controller-gen locally if necessary.
-	$(call go-install-tool,$(CONTROLLER_GEN),sigs.k8s.io/controller-tools/cmd/controller-gen@v0.14.0)
+	$(call go-install-tool,$(CONTROLLER_GEN),sigs.k8s.io/controller-tools/cmd/controller-gen@v0.18.0)
```
SchSeba, line 148: you need to run `make -f Makefile.bundle bundle` after this version change

SchSeba (commented):
I think for working on this I left a few comments

go.mod
kunalmemane: Not really needed to bump it to newer major, we can keep it to the same minor as go directive.

pkg/host/internal/kernel/kernel.go
kunalmemane: I will update it as per the main and u/s.

pkg/host/internal/sriov/sriov.go
kunalmemane: Yes, It is same as main and u/s.

Makefile
kunalmemane: will do. Thanks!

SchSeba (commented):
/lgtm
/approve

PR #1229, opened 2026-05-26
Fix metrics exporter test: replace wget with curl

sshnaidm:
The metrics exporter container image does not include wget, causing the "collects metrics regarding receiving traffic via VF" test to fail. Use curl instead, which is consistent with the same file's Prometheus query helper.


SchSeba:
/hold

it was added here https://github.com/openshift/sriov-network-operator/pull/1220/changes/69a48f10bed18547e2f3a203b93b582528e919b2 from u/s because the image didn't contain curl

SchSeba:
I also check the metrics exporter image and it does contains wget
```
wget-1.21.1-8.el9_4
```

PR #1244, opened 2026-07-05
rbac: grant operator access to config.openshift.io apiservers

zeeke:
Leftover from
- https://github.com/openshift/sriov-network-operator/pull/1239

cc @SchSeba


SchSeba (approved):
/lgtm
/approved
/verified by ci

PR #1245, opened 2026-07-10
Add SamD2021, pablmart, and zhiqiangf to OWNERS

wizhaoredhat:
## Summary
- Add SamD2021, pablmart, and zhiqiangf to OWNERS as reviewers and approvers


SchSeba:
Hi @wizhaoredhat if we're already adding new folks to the list can you also please do a cleanup?
thanks!

